PR #29, opened 2016-01-08
Fix Python Bindings for CYGWIN

long-m-r:
Checks if the 'posix' os.name is actually CYGWIN. If it is, then treat the system as Windows.

Fix allows the package to be imported into a CYGWIN python distribution and allows for the creation of empty Indigo() object without any errors. I have not tested any other features so far.


long-m-r:
Updated further files in same manner. Decided to delete this branch and re-submit with all files

PR #30, opened 2016-01-08
Fix Python Bindings for Cygwin

long-m-r:
Enable correct detection of Windows/NT for Cygwin Python distributions.

Since os.name=='posix' for cygwin, it was incorrectly being detected as linux. This check ensures that Cygwin is treated as windows in Python. It should have no impact on any other system.

Tested by creating an empty Indigo() object without any errors. I did not test further functionality.

### DIFF
--- a/api/plugins/bingo/python/bingo.py
+++ b/api/plugins/bingo/python/bingo.py
@@ -100,9 +100,9 @@ class Bingo(object):
 
     @staticmethod
     def _getLib(indigo):
-        if os.name == 'posix' and not platform.mac_ver()[0]:
+        if os.name == 'posix' and not platform.mac_ver()[0] and not platform.system().startswith("CYGWIN"):
             _lib = CDLL(indigo.dllpath + "/libbingo.so")
-        elif os.name == 'nt':
+        elif os.name == 'nt' or platform.system().startswith("CYGWIN"):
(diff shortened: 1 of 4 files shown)


AlexanderSavelyev:
Dear long-m-r,

You have created a pull request for Indigo repository. I wanted to send by e-mail, but unfortunately, it is not available in your profile

We are very glad to receive this fix from you. It was long time since the PR was created because we have discussed as this PR can be merged internally. The reason is that Indigo project has two licenses (GPL and commercial)

http://lifescience.opensource.epam.com/indigo/index.html#commercial-availability

And by merging your request we have to create a new branch for public contributions. I just want to ask you: are you confirm to give your code by using with this two licences (it means you give the code and agree that we will use it in our product for any need and we will not create a separate branch)? Please let us know if it is not a case and you only agree to give the code under GPL terms. We have to ask you in order to follow "bureaucracy"  rules

Again thank you very much for your participating our project, it is very important for us

With best regards
Alexander Savelev
Indigo project Team Lead

AlexanderSavelyev:
@long-m-r  could you let us know about the question above?

long-m-r:
Sure, that isn't a problem. Sorry for the late response, missed the first
email.

-Matt

Issue #50, opened 2016-05-05
Fix bingo postgres build procedure


AlexanderSavelyev:
CMake pack seems to be configured incorrectly for Bingo PostgreSQL
Please see 
https://groups.google.com/forum/#!topic/indigo-general/w_LRX109pfY
for some details

PR #51, opened 2016-05-05
Fixed bingo SQL installl scripts path after #50

Lucas-Gluchowski:
#50 introduced Postgres versioning of build but installation SQL scripts hasn't been changed.
I checked it on Mac OS X, so I might be worth checking it on Windows.

### DIFF
--- a/bingo/postgres/bingo-pg-install.bat
+++ b/bingo/postgres/bingo-pg-install.bat
@@ -105,15 +105,15 @@ echo     Set OutStream = FS.OpenTextFile(FileName, 8, True) >> replace.vbs
 echo     OutStream.Write Contents >> replace.vbs
 echo End Function >> replace.vbs
 @rem Generate install script
-cscript //b replace.vbs BINGO_SCHEMANAME %schemaname% sql\bingo_schema.sql.in bingo_install.sql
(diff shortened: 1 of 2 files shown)


AlexanderSavelyev:
@Lucas-Gluchowski did you specify environment variables?
BINGO_PG_DIR and BINGO_PG_VERSION 

as it is required here
https://github.com/epam/Indigo/blob/master/bingo/postgres/README-BUILD.md

Lucas-Gluchowski:
@AlexanderSavelyev I have added env variables BINGO_PG_DIR and BINGO_PG_VERSION but only when I added /common/ to path building script worked.Take a look into directory structure (https://github.com/epam/Indigo/tree/master/bingo/postgres/sql) there are 9.0, 9.1, 9.2, 9.3, 9.4, 9.5 and common directories but last is not included in path building.

AlexanderSavelyev:
@Lucas-Gluchowski please check the "dist" folder after the build procedure is finished. It should contain an archive, e.g. bingo-postgres9.4-1.7.10-dev7-linux64.zip
This a deliverable bingo cartridge, which should be used to install for end user. The idea is that sql folder inside does not contain "common" or any other subfolders. In other words, you should not use the "bingo/postgres/bingo-pg-install.sh" script directly

Lucas-Gluchowski:
Looks like I misuse installation scripts. I didn't use final cartridge, but produced SQL scripts.

PR #105, opened 2017-11-15
Chem similarity epmlsopket 1083 oct17

voddan:


voddan:
Add a way for bingo to switch between "sim" and "chem" fingerprints.

Implementation options:
- A new bingo option to switch `()::bingo.sim` to use "chem"
- A new operator akin to `select * from $table where $column @ ($bottom, $top, '$query', '$metric')::bingo.sim`

voddan:
Turn off the "chem" fingerprint by default. That includes not using it in functions (like fingerprint("full")) and not building bytes for it.

PR #192, opened 2020-10-01
map c++ method of one bits list into java

### LINKED ISSUE #191: oneBitsList method is missing in Java implementation
Apparently it hasn't been mapped into Java API

MysterionRise:
Should fix #191

### DIFF
--- a/api/java/src/test/java/com/epam/indigo/IndigoTests.java
+++ b/api/java/src/test/java/com/epam/indigo/IndigoTests.java
@@ -19,4 +19,16 @@ public class IndigoTests {
                 indigoObject.canonicalSmiles(),
                 "C1=CC=CC=C1 is the same as C1C=CC=CC=1");
     }
+
+    @Test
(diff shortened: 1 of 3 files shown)


api/java/src/test/java/com/epam/indigo/IndigoTests.java
```diff
@@ -19,4 +19,16 @@ void loadMoleculeFromSMILES() {
...
+    @Test
+    @DisplayName("Loading molecule and getting one bits list")
+    void getOneBitsList() {
+        Indigo indigo = new Indigo();
+        IndigoObject indigoObject = indigo.loadMolecule("C1=CC=CC=C1");
+
+        assertEquals(
+                "1698 1719 1749 1806 1909 1914 1971 2056",
+                indigoObject.fingerprint().oneBitsList(),
+                "same one bits as in string 1698 1719 1749 1806 1909 1914 1971 205");
```
SunFellow, line 32: Why missed the last `6` in the string? (`205` instead `2056`)

PR #201, opened 2020-10-25
deprecate unserialize in python,dotnet,java

MysterionRise:


khyurri:
Could you please format api/python/indigo.py using:

```
$ black api/python/indigo.py --line-length=79
$ isort api/python/indigo.py
```

If you have not yet installed black and isort, both are available in pip.

PR #206, opened 2020-10-30
api: tests: added python functional tests

mkviatkovskii:


api/tests/python/README.md
```diff
@@ -0,0 +1,63 @@
+# Тестирование Indigo
```
khyurri, line 1: Should we create issue to translate this Readme?

Issue #180, opened 2020-09-02
Update Implementation of JSON data format to support Ketcher's version


mkviatkovskii:

AlexanderSavelyev:
See https://github.com/epam/Indigo/wiki/JSON-structure-format

SunFellow:
Related Ketcher issue: https://github.com/epam/ketcher/issues/53
Related Ketcher pull request: https://github.com/epam/ketcher/pull/55

PR #224, opened 2020-11-18
WIP: core: Ketcher format support (#180)

mkviatkovskii:


api/indigo.h
```diff
@@ -208,7 +212,7 @@ CEXPORT const char* indigoCdxml(int object);
 // (because MDLCT data usually contains zeroes)
 CEXPORT int indigoSaveMDLCT(int item, int output);
 
-CEXPORT const char* indigoJson(int object);
+// CEXPORT const char* indigoJson(int object);
```
SunFellow, line 215: Better remove this function completely along with its implementation.

api/src/indigo_misc.cpp
```diff
@@ -1334,6 +1334,7 @@ CEXPORT const char* indigoCheckStructure(const char* structure, const char* prop
     INDIGO_END(0);
 }
 
+/*
 CEXPORT const char* indigoJson(int item)
```
SunFellow, line 1338: Better remove this function completely for clarity, not to lid the code with commented stuff.

api/src/indigo_internal.h
```diff
@@ -82,6 +83,10 @@ class DLLEXPORT IndigoObject
         REACTION_ITER,
         REACTION_MOLECULE,
         GROSS_MOLECULE,
+        JSON_LOADER,
```
SunFellow, line 86: IMHO,`JSON_LOADER` and `JSON_SAVER` and corresponding classes, if referring to loader and saver functionality, do not have place in the model of indigoObject types hierarchy. 
Note: The same is true regarding to `SDF_LOADER` and `SDF_SAVER` and corresponding classes, but they are out of scope of this change.

Anyway, once you rejected `class IndigoJSONLoader` inherited from `class IndigoObject`, these enums should be rejected too.

api/src/indigo_internal.h
```diff
@@ -82,6 +83,10 @@ class DLLEXPORT IndigoObject
         REACTION_ITER,
         REACTION_MOLECULE,
         GROSS_MOLECULE,
+        JSON_LOADER,
+        JSON_SAVER,
```
SunFellow, line 87: IMHO,`JSON_LOADER` and `JSON_SAVER` and corresponding classes, if referring to loader and saver functionality, do not have place in the model of indigoObject types hierarchy. 
Note: The same is true regarding to `SDF_LOADER` and `SDF_SAVER` and corresponding classes, but they are out of scope of this change.

Anyway, once you rejected `class IndigoJSONLoader` inherited from `class IndigoObject`, these enums should be rejected too.

api/src/indigo_loaders.cpp
```diff
@@ -52,6 +89,35 @@ IndigoSdfLoader::~IndigoSdfLoader()
...
+bool IndigoSdfLoader::hasNext()
+{
+    return !sdf_loader->isEOF();
+}
+
+long long IndigoSdfLoader::tell()
+{
+    return sdf_loader->tell();
+}
+
```
SunFellow, line 120: Indeed these methods of `IndigoSdfLoader` were not implemented? 
Anyway, this is a fix and it does not have relation to the subject of the pull request (support Json for Ketcher). These changes should be extracted to a separate pull request and issue.

api/src/indigo_loaders.h
```diff
@@ -121,6 +136,24 @@ class IndigoRdfLoader : public IndigoObject
     AutoPtr<Scanner> _own_scanner;
 };
 
+
+class IndigoJSONMolecule : public IndigoObject
```
SunFellow, line 140: Are you sure to inherit from `IndigoObject`? Maybe from `IndigoMolecule`, which already incapsulates `Molecule`?

api/src/indigo_internal.h
AlexanderSavelyev: Don't quite get the comment. Why the SDF_LOADER and SDF_SAVER (and same for JSON) are not part of IndigoObject?

api/src/indigo_loaders.h
AlexanderSavelyev: JSON Molecule is not a Molecule (e.g. JSON object can be SDF), so it is worth keep it separate

api/src/indigo_internal.h
SunFellow: I would qualify them in another branch of hierarchy not to mix with objects->molecules->etc. 
But, yeah, now I see that you built yet another OO model, where everything is derived from IndigoObject. 
So forget the generic part of the comment.
Just we have to maintain correspondence between the enums and the existing Object classes. No class - no enum.

api/src/indigo_loaders.h
SunFellow: BTW, why rejected the initial approach, when a loader was an operation on a molecule?
Introducing JSON Molecule, SDF Molecule, we bind domain objects to the format of storage. 
They are completely different objects and cannot be simply casted each other (I mean JSONModlecule to SDD Molecule or the like), we'll have to add special converters.

SunFellow (changes requested):
@mkviatkovskii 
Please pay attention to  `api/plugins/bingo-elastic/java/src/test/resources/zinc-slice.sdf.gz`
which hit to this commit eb2951b125e4a8970334624b4402a329e7b14352 as modified.
I believe it should not be changed here. I guess it is caused by the following factors:
1) After LFS support was added, `.gz` file were defined as Text files (see `.gitattributes ` at the root folder). (Probably this file was added in parallel with LFS support and was not converted to an URL pointing to the binary object in LFS store.)
2) The usual git config sets AutoCRLF=true.
This results in replacing instances of bytes LF byte to CR LF pair of bytes in text files (including this .gz) when checking out and git goes crazy: it thinks it is always modified. I fight with this at my workspace for a long time until I disable this CR/LF conversion.

Please exclude it from the commit And we need to think what to do with those .gz files.
P.S. I also noticed that you've clean up this file to empty on another branch.

api/src/indigo_loaders.cpp
```diff
@@ -469,11 +542,14 @@ IndigoObject* IndigoMultilineSmilesLoader::at(int index)
...
-}
-INDIGO_END(-1)
+CEXPORT int indigoIterateSDF(int reader)
+{
+    INDIGO_BEGIN
+    {
+        IndigoObject& obj = self.getObject(reader);
+        return self.addObject(new IndigoSdfLoader(IndigoScanner::get(obj)));
+    }
+    INDIGO_END(-1)
```
SunFellow, line 552: It turned out that we need to put semicolon after `INDIGO_END()` macro. Refer to issue #267.

api/src/indigo_savers.cpp
```diff
@@ -526,7 +527,31 @@ return 1;
...
+            return 1;
+        }
+        if (IndigoBaseReaction::is(obj))
+        {
+            // reaction not implemented yet
+            return 1;
+        }
+        throw IndigoError("indigoSaveCml(): expected molecule or reaction, got %s", obj.debugInfo());
+    }
+    INDIGO_END(-1)
```
SunFellow, line 549: Need to put semicolon after `INDIGO_END()` macro.

molecule/base_molecule.h
```diff
@@ -306,6 +312,21 @@ namespace indigo
...
+		void selectAtoms(const Filter& filter);
+		void selectBonds(const Filter& filter);
+		void unselectAtom(int idx);
+		void unselectBond(int idx);
+		int countSelectedAtoms();
+		int countSelectedBonds();
+		bool hasSelection();
+		bool isAtomSelected(int idx);
+		bool isBondSelected(int idx);
+		void selectSubmolecule(BaseMolecule& sub, const int* mapping, bool entire);
```
SunFellow, line 327: Why there are tabs instead spaces?
This happens with this commit: fe9aefb91d8e5e7bb7448bda50fe196f61774591 on 12.01.2020.
Probably, clang autoformatting was disabled?

molecule/base_molecule.h
```diff
@@ -383,6 +404,9 @@ namespace indigo
 
         Array<int> _hl_atoms;
         Array<int> _hl_bonds;
+		Array<int> _sl_atoms;
+		Array<int> _sl_bonds;
```
SunFellow, line 408: Here are tabs too.

molecule/molecule_json_loader.h
```diff
@@ -43,13 +47,28 @@ namespace indigo
...
+        void loadMolecule( BaseMolecule& mol );
+        int addAtomToMoleculeQuery( const char* label, int element, int charge, int valence, int radical, int isotope );
+        int addBondToMoleculeQuery( int beg, int end, int order, int topology = 0 );
+        void validateMoleculeBond( int order );
+        void parseAtoms( const rapidjson::Value& atoms, BaseMolecule& mol );
+        void parseBonds( const rapidjson::Value& bonds, BaseMolecule& mol, int atom_base_idx );
+		void parseHighlightedBonds( const rapidjson::Value& bonds, BaseMolecule& mol );
+		void parseHighlightedAtoms(const rapidjson::Value& atoms, BaseMolecule& mol);
+		void parseSelectedBonds(const rapidjson::Value& bonds, BaseMolecule& mol);
+		void parseSelectedAtoms(const rapidjson::Value& atoms, BaseMolecule& mol);
```
SunFellow, line 60: Tabs.

molecule/molecule_json_loader.h
```diff
@@ -32,8 +34,10 @@ namespace indigo
 {
 
     class Scanner;
```
SunFellow, line 36: Do we still need `Scanner` here?

molecule/molecule_json_saver.h
```diff
@@ -39,15 +39,25 @@ namespace indigo
...
+        void saveMolecule( BaseMolecule& bmol );
+        void saveAtoms( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveBonds( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveRGroup( PtrPool<BaseMolecule>& fragments, int rgnum, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+		void saveSGroups( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+		void saveSGroup(SGroup& sgroup, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+		void saveAttachmentPoint(BaseMolecule& mol, int atom_idx, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+		void saveStereoCenter(BaseMolecule& mol, int atom_idx, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+		void saveHighlights(BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+		void saveSelection(BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer);
```
SunFellow, line 52: Tabs.

molecule/molecule_json_saver.h
```diff
@@ -39,15 +39,25 @@ namespace indigo
...
 
         DECL_ERROR;
 
     protected:
-        Molecule* _mol;
+		void _checkSGroupIndices(BaseMolecule& mol, Array<int>& sgs_list);
```
SunFellow, line 57: Tabs.

molecule/src/molecule_json_loader.cpp
```diff
@@ -2,67 +2,280 @@
...
+				strscan.skipSpace();
+				strscan.readWord(id, 0);
+				pyramid[k] = getAtomIdx(id.ptr());
+				if (pyramid[k] == i)
+					pyramid[k] = -1;
+			}
+
+			MoleculeStereocenters::moveMinimalToEnd(pyramid);
+			mol.stereocenters.add(i, MoleculeStereocenters::ATOM_ABS, 0, pyramid);*/
+		}
```
SunFellow, line 201: Tabs.

molecule/src/molecule_json_loader.cpp
```diff
@@ -71,33 +284,370 @@ void MoleculeJsonLoader::loadMolecule(Molecule& mol)
...
+	}
+}
+
+void indigo::MoleculeJsonLoader::parseSelectedAtoms(const rapidjson::Value & atoms, BaseMolecule & mol)
+{
+	for (int i = 0; i < atoms.Size(); ++i)
+	{
+		mol.selectBond(atoms[i].GetInt());
+	}
+}
```
SunFellow, line 383: Tabs.

molecule/src/molecule_json_loader.cpp
```diff
@@ -71,33 +284,370 @@ void MoleculeJsonLoader::loadMolecule(Molecule& mol)
...
+				parseHighlightedAtoms(mol_node["hl_atoms"], mol);
+			}
+			if (mol_node.HasMember("sl_bonds"))
+			{
+				parseSelectedBonds(mol_node["sl_bonds"], mol);
+			}
+			if (mol_node.HasMember("sl_atoms"))
+			{
+				parseSelectedAtoms(mol_node["sl_atoms"], mol);
+			}
```
SunFellow, line 614: Tabs.

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
-    std::stringstream result;
-    QS_DEF(Array<char>, buf);
-    ArrayOutput out(buf);
+	QS_DEF(Array<int>, sgs_sorted);
+    _checkSGroupIndices(mol, sgs_sorted);
 
-    _mol = &mol;
+    if (mol.countSGroups() > 0)
+    {
+		writer.Key("sgroups");
```
SunFellow, line 140: Tabs.

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+        writer.StartArray();
+        int idx = 1;
+        for (int i = 0; i < sgs_sorted.size(); i++)
+        {
+            int sg_idx = sgs_sorted[i];
+			saveSGroup( mol.sgroups.getSGroup(sg_idx), writer );
+		}
+		writer.EndArray();
+	}
```
SunFellow, line 149: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+	writer.Key("type");
+	writer.String(SGroup::typeToString(sgroup.sgroup_type));
+	writer.Key("atoms");
+	if (sgroup.sgroup_type != SGroup::SG_TYPE_MUL)
+	{
+		writer.StartArray();
+		for (int i = 0; i < sgroup.atoms.size(); i++)
+			writer.Int(sgroup.atoms[i]);
+		writer.EndArray();
+	}
```
SunFellow, line 164: Tabs.

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+			throw Error("SG_TYPE_FOR not implemented in indigo yet");
+		break;
+		case SGroup::SG_TYPE_ANY:
+			throw Error("SG_TYPE_ANY not implemented in indigo yet");
+		break;
+		default:
+		break;
+	}
+	writer.EndObject();
+}
```
SunFellow, line 321: Tabs used for indent.

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+
+void MoleculeJsonSaver::saveBonds( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer )
+{
+    QS_DEF(Array<char>, buf);
+    ArrayOutput out(buf);
+    if (mol.edgeCount() > 0)
     {
-        writer.Key("atoms");
+		writer.Key("bonds");
```
SunFellow, line 329: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
         writer.StartArray();
-        for (auto i : _mol->vertices())
+        for (auto i : mol.edges())
         {
             writer.StartObject();
+            writer.Key("type");
+			int bond_order = mol.getBondOrder(i);
+			if (bond_order < 0 && _pqmol )
```
SunFellow, line 336: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+                else if (qb == QueryMolecule::QUERY_BOND_ANY)
+                    bond_order = 8;
+			   if( bond_order < 0 )
+                 throw Error("Invalid query bond");
+            }
+
+            if (bond_order == BOND_ZERO && _pmol )
+            {
+                bond_order = 9;
+				const Edge& edge = mol.getEdge(i);
```
SunFellow, line 354: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+			if (_pqmol)
+			{
+				_pqmol->getBond(i).sureValue(QueryMolecule::BOND_TOPOLOGY, topology);
+				if ( topology > 0)
+				{
+					writer.Key("topology");
+					writer.Uint(topology);
+				}
+			}
+
```
SunFellow, line 371: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
-            if (have_hyz)
+            if( stereo )
             {
-                const Vec3f& coord = _mol->getAtomXyz(i);
+                writer.Key("stereo");
+                writer.Uint( stereo );
+            }
+            writer.EndObject();
+        }
+		writer.EndArray();
```
SunFellow, line 415: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+				break;
+			}
+		}
+	}
+
+	if (val > 0)
+	{
+		writer.Key("attachmentPoints");
+        writer.Int(val);
+	}
```
SunFellow, line 439: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+	QS_DEF(Array<char>, buf);
+	ArrayOutput out(buf);
+	const int* pyramid = mol.stereocenters.getPyramid(atom_idx);
+	if (pyramid[3] == -1)
+		out.printf("a%d a%d a%d a%d", pyramid[0], pyramid[1], pyramid[2], atom_idx);
+	else
+		out.printf("a%d a%d a%d a%d", pyramid[0], pyramid[1], pyramid[2], pyramid[3]);
+	buf.push(0);
+	writer.Key("atomRefs4");
+	writer.String(buf.ptr());
```
SunFellow, line 453: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+		writer.StartArray();
+		for (int i = mol.vertexBegin(); i != mol.vertexEnd(); i = mol.vertexNext(i))
+		{
+			if (mol.isAtomHighlighted(i))
+			{
+				writer.Int(i);
+			}
+		}
+		writer.EndArray();
+	}
```
SunFellow, line 483: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+		writer.StartArray();
+		for (int i = mol.vertexBegin(); i != mol.vertexEnd(); i = mol.vertexNext(i))
+		{
+			if (mol.isAtomSelected(i))
+			{
+				writer.Int(i);
+			}
+		}
+		writer.EndArray();
+	}
```
SunFellow, line 513: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+    {
+        for (auto i : mol.vertices())
+        {
+			int anum = mol.getAtomNumber(i);
+			int isotope = mol.getAtomIsotope(i);
+            writer.StartObject();
+			if (mol.attachmentPointCount())
+				saveAttachmentPoint(mol, i, writer);
+			if (mol.stereocenters.getType(i) > MoleculeStereocenters::ATOM_ANY)
+				saveStereoCenter( mol, i, writer );
```
SunFellow, line 532: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+				{
+					if (isotope == 2)
+					{
+						buf.clear(); buf.appendString("D", true);
+					}
+					if (isotope == 3)
+					{
+						buf.clear(); buf.appendString("T", true);
+					}
+				}
```
SunFellow, line 567: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+		mol.reset(new QueryMolecule());
+		_pqmol = (QueryMolecule*) mol.get();
+	} else
+	{
+		mol.reset( new Molecule());
+		_pmol = (Molecule*)mol.get();
+	}
+	mol->clone_KeepIndices( bmol );
+	BaseMolecule::collapse( *mol );
+	QS_DEF(Array<char>, buf);
```
SunFellow, line 660: Tabs

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,707 @@
...
+    writer.Key("atoms");
+    writer.StartArray();
+    saveAtoms( *mol, writer );
+    writer.EndArray();
+
+    saveBonds( *mol, writer );
+	saveSGroups( *mol, writer );
+	saveHighlights( *mol, writer );
+	saveSelection( *mol, writer );
+
```
SunFellow, line 710: Tabs

reaction/reaction_json_loader.h
```diff
@@ -0,0 +1,68 @@
...
+#include "molecule/molecule.h"
+
+namespace indigo
+{
+    
+    class Scanner;
+    class BaseReaction;
+    class Reaction;
+    class QueryReaction;
+	class QueryMolecule;
```
SunFellow, line 35: Tabs

reaction/reaction_json_loader.h
```diff
@@ -0,0 +1,68 @@
...
+        rapidjson::Value& _pluses;
+        rapidjson::Value& _arrows;
+        
+    private:
+        ReactionJsonLoader(const ReactionJsonLoader&); // no implicit copy
+        Reaction* _prxn;
+        QueryReaction* _pqrxn;
+		Molecule _mol;
+		QueryMolecule _qmol;
+		BaseMolecule* _pmol;
```
SunFellow, line 61: Tabs

reaction/reaction_json_loader.h
```diff
@@ -0,0 +1,68 @@
...
+
+    };
+    
+}; // namespace indigo
```
SunFellow, line 65: No need semicolon after namespace.

reaction/src/reaction_json_loader.cpp
```diff
@@ -0,0 +1,55 @@
...
+	{
+		_pmol = &_mol;
+		loader.loadMolecule(_mol);
+	}
+	else if (_pqrxn)
+	{
+		loader.loadMolecule(_qmol);
+		_pmol = &_qmol;
+	} else
+		throw Error("unknown reaction type: %s", typeid(rxn).name());
```
SunFellow, line 53: Tabs.

SunFellow (changes requested):
Please!!! Regarding these DLLs, make "combine the last two to one commit" to annihilate adding binary files and **force** push the branch. Otherwise these DLLS will increase the size of the repository because they stay in the history forever (Github stores all commits bound to a pull request.)

SunFellow:
Done:
1) Cut off the accidently modified binary api/plugins/bingo-elastic/java/src/test/resources/zinc-slice.sdf.gz from the commit on 01/03/2021
2) Cut off the accidentally added and removed DLLs in commits on 01/25/2021.

api/src/indigo_savers.cpp
even1024: fixed

molecule/base_molecule.h
even1024: fixed

molecule/base_molecule.h
even1024: fixed

api/indigo.h
even1024: fixed

api/src/indigo_loaders.cpp
```diff
@@ -36,6 +39,40 @@
...
+    if( json_loader->currentNumber() >= index ) return NULL;
+    return new IndigoJSONMolecule( json_loader->at( index ), json_loader->rgroups(), json_loader->currentNumber() );
+}
+
+bool IndigoJSONLoader::hasNext()
+{
+    return json_loader->hasNext();
+}
+*/
+
```
mkviatkovskii, line 75: Do we need this commented code here?

api/src/indigo_loaders.h
```diff
@@ -87,20 +89,33 @@ class IndigoSdfLoader : public IndigoObject
...
+    virtual IndigoObject* next();
+    virtual bool hasNext();
+    IndigoObject* at(int index);
+    AutoPtr<JSONLoader> json_loader;
+    
+protected:
+    AutoPtr<Scanner> _own_scanner;
+};*/
+
+
```
mkviatkovskii, line 118: Again, is this required?

layout/src/molecule_layout.cpp
```diff
@@ -412,10 +412,18 @@ BaseMolecule& MoleculeLayout::_getMol(int id)
...
 }
+
+void MoleculeLayout::updateSGroups()
+{
+    _updateDataSGroups();
+    _updateMultipleGroups();
+    _updateRepeatingUnits();
+}
+
+
```
mkviatkovskii, line 423: extra empty line

molecule/base_molecule.h
```diff
@@ -57,9 +57,15 @@ namespace indigo
...
+        BOND_AROMATIC = 4,
+        _BOND_SINGLE_OR_DOUBLE = 5,
+        _BOND_SINGLE_OR_AROMATIC = 6,
+        _BOND_DOUBLE_OR_AROMATIC = 7,
+        _BOND_ANY = 8,
+        _BOND_COORDINATION = 9,
+        _BOND_HYDROGEN = 10
     };
-
+
```
mkviatkovskii, line 68: extra spaces

molecule/molecule_json_saver.h
```diff
@@ -23,11 +23,11 @@
 
 #include "third_party/rapidjson/stringbuffer.h"
 #include "third_party/rapidjson/writer.h"
-
+#include "molecule/base_molecule.h"
 #include "base_cpp/locale_guard.h"
 #include "base_cpp/output.h"
 #include "molecule/elements.h"
-#include "molecule/molecule.h"
+#include <set>
```
mkviatkovskii, line 30: Why do we need exactly `<set>`?

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,771 @@
  ***************************************************************************/
 
 #include "molecule/molecule_json_saver.h"
+#include "molecule/molecule.h"
+#include "molecule/query_molecule.h"
+#include "layout/molecule_layout.h"
+
+#include <vector>
+#include <memory>
+#include <set>
```
mkviatkovskii, line 26: Why do we need `<set>`?

mkviatkovskii (commented):
I suggest running clang-format on all code and take a look at my comments.

api/src/indigo_loaders.h
```diff
@@ -24,6 +24,8 @@
 #include "base_cpp/properties_map.h"
 #include "molecule/molecule.h"
 #include "reaction/reaction.h"
+#include "molecule/query_molecule.h"
+#include "third_party/rapidjson/document.h"
```
SunFellow, line 28: It's better to provide `third_party` folder in includes using project's settings, and leave only 
```
#include "rapidjson/document.h"
```
or even better (since this is third-party libraries):
```
#include <rapidjson/document.h>
```
This would make the code independent from possible further changes in approach to attach third-party libs to the project - you would need to change it in one place.

api/src/indigo_loaders.cpp
```diff
@@ -91,6 +157,45 @@ int IndigoRdfData::getIndex()
...
+{
+    if (getMolecule().name.ptr() == 0)
+        return "";
+    return getMolecule().name.ptr();
+}
+
+IndigoJSONMolecule::~IndigoJSONMolecule()
+{
+}
+
```
SunFellow, line 198: It's hard to identify among a large pull-request that these strings are just moved. 
Maybe keep them in-place?

api/src/indigo_loaders.cpp
SunFellow: And finally, it turns out that this file has no essential change.

api/src/indigo_internal.h
```diff
@@ -56,6 +56,7 @@ namespace indigo
     class Output;
     class Scanner;
     class SdfLoader;
+    class JSONLoader;
```
SunFellow, line 59: Is there any usage of `class JSONLoader`?

api/src/indigo_object.cpp
```diff
@@ -122,6 +122,10 @@ IndigoObjectTypes::IndigoObjectTypes()
     emplace(IndigoObject::TGROUP, "TGroup");
     emplace(IndigoObject::TGROUPS_ITER, "TGroupsIterator");
     emplace(IndigoObject::GROSS_REACTION, "GrossReaction");
+    emplace(IndigoObject::JSON_LOADER, "JsonLoader");
+    emplace(IndigoObject::JSON_SAVER, "JsonSaver");
```
SunFellow, line 126: Do we need `JSON_LOADER` and `JSON_SAVER` if the corresponding classes are not used?

api/src/indigo_internal.h
```diff
@@ -159,6 +160,10 @@ class DLLEXPORT IndigoObject
         TGROUP,
         TGROUPS_ITER,
         GROSS_REACTION,
+        JSON_LOADER,
+        JSON_SAVER,
```
SunFellow, line 164: Do we need JSON_LOADER and JSON_SAVER if the corresponding classes are not used?

api/src/indigo_savers.cpp
```diff
@@ -545,21 +546,43 @@ CEXPORT int indigoSaveMolfile(int molecule, int output)
...
+            saver.saveMolecule( mol );
+            out.flush();
+            return 1;
+        }
+        if (IndigoBaseReaction::is(obj))
+        {
+            // reaction not implemented yet
+            return 1;
+        }
+        throw IndigoError("indigoSaveCml(): expected molecule or reaction, got %s", obj.debugInfo());
```
SunFellow, line 575: `IndigoError("indigoSaveCml()` in function `indigoSaveJson` - copy-paste error.

api/indigo.h
```diff
@@ -807,6 +811,7 @@ CEXPORT int indigoIterateRDF(int reader);
 CEXPORT int indigoIterateSmiles(int reader);
 CEXPORT int indigoIterateCML(int reader);
 CEXPORT int indigoIterateCDX(int reader);
+CEXPORT int indigoIterateJSON(int reader);
```
SunFellow, line 814: I can't find definition of this function.

molecule/molecule_json_loader.h
```diff
@@ -43,13 +48,26 @@ namespace indigo
...
+        void loadMolecule( BaseMolecule& mol );
+        int addAtomToMoleculeQuery( const char* label, int element, int charge, int valence, int radical, int isotope );
+        int addBondToMoleculeQuery( int beg, int end, int order, int topology = 0 );
+        void validateMoleculeBond( int order );
+        void parseAtoms( const rapidjson::Value& atoms, BaseMolecule& mol );
+        void parseBonds( const rapidjson::Value& bonds, BaseMolecule& mol, int atom_base_idx );
+        void parseHighlight( const rapidjson::Value& highlight, BaseMolecule& mol );
+        void parseSelection(const rapidjson::Value& selection, BaseMolecule& mol);
+        void parseSGroups( const rapidjson::Value& sgroups, BaseMolecule& mol );
+        void handleSGroup(SGroup& sgroup, const std::unordered_set<int>& atoms, BaseMolecule& bmol);
```
SunFellow, line 61: I'd make these auxiliary methods `protected`.

molecule/molecule_json_saver.h
```diff
@@ -39,15 +38,25 @@ namespace indigo
...
+        void saveMolecule( BaseMolecule& bmol );
+        void saveAtoms( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveBonds( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveRGroup( PtrPool<BaseMolecule>& fragments, int rgnum, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveSGroups( BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer );
+        void saveSGroup(SGroup& sgroup, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+        void saveAttachmentPoint(BaseMolecule& mol, int atom_idx, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+        void saveStereoCenter(BaseMolecule& mol, int atom_idx, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+        void saveHighlights(BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer);
+        void saveSelection(BaseMolecule& mol, rapidjson::Writer<rapidjson::StringBuffer>& writer);
```
SunFellow, line 51: I'd make these auxiliary methods protected.

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,771 @@
...
+            if (orig_ids.find(sgroup.parent_group) == -1 || sgroup.parent_group == sgroup.original_group)
+            {
+                sgroup.parent_group = 0;
+                sgs_list.push(i);
+                added_ids.push(sgroup.original_group);
+            }
+        }
+    }
+
+    for (;;)
```
SunFellow, line 113: Infinite cycle and it seems no break from it?

molecule/src/molecule_json_saver.cpp
```diff
@@ -17,116 +17,771 @@
...
+
+            if (added_ids.find(sgroup.parent_group) != -1)
+            {
+                sgs_list.push(i);
+                added_ids.push(sgroup.original_group);
+            }
+        }
+        if (sgs_list.size() == mol.countSGroups())
+            break;
+    }
```
SunFellow, line 132: Please note that this `break;` is out from the inner `for`:
```
for (int i = mol.sgroups.begin(); i != mol.sgroups.end(); i = mol.sgroups.next(i))
```
The outer ` for (;;)` does not have a condition to end.

api/src/indigo_savers.cpp
```diff
@@ -545,21 +546,43 @@ CEXPORT int indigoSaveMolfile(int molecule, int output)
...
+            MoleculeJsonSaver saver(out);
+            BaseMolecule& mol = obj.getBaseMolecule();
+            saver.saveMolecule( mol );
+            out.flush();
+            return 1;
+        }
+        if (IndigoBaseReaction::is(obj))
+        {
+            // reaction not implemented yet
+            return 1;
```
SunFellow, line 573: Maybe `-1` until not implemented?

api/tests/python/tests/formats/molecules/cis_trans.ket.json
```diff
@@ -0,0 +1,69 @@
+{
```
SunFellow, line 1: This file is similar to `cis_trans.ket`. Do we need to keep them both?

api/tests/python/tests/formats/molecules/cis_transi.ket
```diff
@@ -0,0 +1 @@
+{"root":{"nodes":[{"$ref":"mol0"}]},"mol0":{"type":"molecule","atoms":[{"label":"F","location":[1.5260000228881837,-1.5499999523162842,0.0]},{"label":"C","location":[2.3919999599456789,-2.049999952316284,0.0]},{"label":"C","location":[3.257999897003174,-1.5499999523162842,0.0]},{"label":"F","location":[4.124000072479248,-2.049999952316284,0.0]}],"bonds":[{"type":1,"atoms":[0,1]},{"type":2,"atoms":[1,2],"stereo":8},{"type":1,"atoms":[2,3]}]}}
```
SunFellow, line 1: This file looks like a version of previous files `cis_trans.ket` or `cis_trans.ket.json` with some rounds, new field `stereo` and compressed. Do we need it?

molecule/src/json_loader.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <string>
+
+using namespace indigo;
+using namespace rapidjson;
+
+IMPL_ERROR(JSONLoader, "JSON loader");
+
+CP_DEF(JSONLoader);
+
+JSONLoader::JSONLoader( Scanner& scanner ) : CP_INIT, _current_number( 0 ), _molecules( kArrayType ), _rgroups( kArrayType )
```
SunFellow, line 32: Seems unused `class JSONLoader`, unused cpp, unused code.

molecule/molecule.h
```diff
@@ -214,6 +214,7 @@ namespace indigo
 
             Array<_AttachOrder> order; // attach order info
         };
+
```
SunFellow, line 217: There is no essential change in this file. It's worse to exclude from the whole file from pr.

molecule/base_molecule.h
```diff
@@ -383,6 +404,9 @@ namespace indigo
 
         Array<int> _hl_atoms;
         Array<int> _hl_bonds;
+        Array<int> _sl_atoms;
+        Array<int> _sl_bonds;
```
SunFellow, line 408: Basically, they should be `Array<bool>`, since selection state can be only: 0/1 or false/true, or unselected/selected.
At least, this would save memory twice. 
The same refers to old `_hl_atoms` and `_hl_bonds`.
Of course, `class Array` is not perfect for `bool` type, but nothing limits us from using `std::vector<bool>` or (differently) `std::set<int>`, where `int` is `index`.

molecule/src/base_molecule.cpp
```diff
@@ -3600,6 +3642,16 @@ void BaseMolecule::unhighlightAtom(int idx)
     }
 }
 
+void BaseMolecule::unselectAtom(int idx)
+{
+	if (_sl_atoms.size() > idx)
+	{
+		_sl_atoms[idx] = 0;
```
SunFellow, line 3649: What if `idx < 0`?

molecule/molecule_json_loader.h
```diff
@@ -43,13 +48,26 @@ namespace indigo
     {
     public:
         DECL_ERROR;
+        explicit MoleculeJsonLoader( rapidjson::Value& molecule, rapidjson::Value& rgroups );
```
SunFellow, line 51: Are you sure that constructor with two rapidjson semi-parsed structures is a handy and enough-abstract way of creating JsonLoader?
I'd load class internals with parsing job, leaving:
```
explicit MoleculeJsonLoader( const char * json_data);
```
Even wasting the previous parsing done by caller. 
BTW and IMHO, there is dangerous operation by re-making those `molecule` and `rgroups` nodes.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+
+        if (a.HasMember("type"))
+        {
+            std::string atom_type = a["type"].GetString();
+            if (atom_type == "rg-label" && a.HasMember("$refs"))
+            {
+                std::string ref = a["$refs"][0].GetString();
+                if (ref.find("rg-") == 0 && ref.erase(0, 3).size())
+                {
+                    rsite_idx = std::stoi(ref);
```
SunFellow, line 207: Some range checking is welcome.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+            else
+            {
+                elem = Element::fromString2(label.c_str());
+                if (elem == -1)
+                {
+                    elem = ELEM_PSEUDO;
+                    if (isotope != 0)
+                    {
+                        throw Error("isotope number not allowed on pseudo-atoms");
+                    }
```
SunFellow, line 239: In the copied-pasted code, `isotope` was set initially:
```
        isotope = _scanner.readIntFix(2);
```
Here `isotope` is initialized by zero (`0`) and is not changed above if hit to this clause.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+        }
+
+        if (a.HasMember("charge"))
+            charge = a["charge"].GetInt();
+        if (a.HasMember("explicitValence"))
+            valence = a["explicitValence"].GetInt();
+        if (a.HasMember("radical"))
+            radical = a["radical"].GetInt();
+        if (a.HasMember("isotope"))
+            isotope = a["isotope"].GetInt();
```
SunFellow, line 251: Here `isotope` is overwritten, no matter it was defined for `D` or `T` atoms.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
 
+        if (_pmol)
+        {
+            atom_idx = _pmol->addAtom(elem);
+            if (charge)
```
SunFellow, line 256: `if (charge)` checks not for the existence, but for the value. Why it cannot get value = `0`?

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+                _pmol->setAtomCharge_Silent(atom_idx, charge);
+            if (valence)
```
SunFellow, line 258: `if (valence)` checks not for the existence, but for the value. 
The schema says:
```
		"explicitValence": {
			"title": "Valence",
			"type": "integer",
			"enum": [ -1,0,1,2,3,4,5,6,7,8,9,10,11,12 ]
		},
```
So `0` is one of legitim values.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+                _pmol->setExplicitValence(atom_idx, valence);
+            if (radical)
```
SunFellow, line 260: Similar comment

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+                _pmol->setAtomRadical(atom_idx, radical);
+            if (isotope)
```
SunFellow, line 262: Similar comment.

molecule/src/molecule_json_loader.cpp
SunFellow, line 251: Wherever `GetInt()` from JSON, some check for allowable values is welcome.

molecule/src/molecule_json_loader.cpp
```diff
@@ -1,68 +1,280 @@
...
+                _pmol->setAtomIsotope(atom_idx, isotope);
+            if (elem == ELEM_PSEUDO)
+            {
+                _pmol->setPseudoAtom(atom_idx, label.c_str());
+            }
+        }
+        else
```
SunFellow, line 269: Nevertheless, using 
``` 
if (_pqmol)
```
would be more proper, than else of `if (_pmol)`...

molecule/src/smiles_loader.cpp
```diff
@@ -1887,6 +1887,7 @@ void SmilesLoader::_handlePolymerRepetition(int i)
         if (_atoms[j].ends_polymer)
             end = j;
     }
+
```
SunFellow, line 1890: This file contains no essential change. It's worth to exclude it from the pr.

molecule/src/molecule_auto_loader.cpp
```diff
@@ -290,24 +289,58 @@ void MoleculeAutoLoader::_loadMolecule(BaseMolecule& mol, bool query)
...
+                            const char* node_name = nodes[i]["$ref"].GetString();
+                            Value& node = data[ node_name ];
+                            std::string node_type = node["type"].GetString();
+                            if( node_type.compare("molecule") == 0 )
+                            {
+                                mol_nodes.PushBack( node, data.GetAllocator() );
+                            } else if (node_type.compare("rgroup") == 0 )
+                            {
+                                rgroups.PushBack( node, data.GetAllocator() );
+                            } else
```
SunFellow, line 324: IMHO, this not a good way to parse JSON and then construct two nodes (`node` and `rgroup`). It can result in losing information from JSON. The ask of auto loader is just to identify the type and call a corresponding loader. The rest of parsing should be done inside the loader.

molecule/src/molecule_auto_loader.cpp
```diff
@@ -290,24 +289,58 @@ void MoleculeAutoLoader::_loadMolecule(BaseMolecule& mol, bool query)
...
-                    else
-                        loader.loadMolecule((Molecule&)mol);
+                    Array<char> buf;
+                    _scanner->readAll(buf);
+                    buf.push(0);
+                    Document data;
+                    Value rgroups(kArrayType);
+                    Value mol_nodes(kArrayType);
+                    if ( data.Parse(buf.ptr()).HasParseError())
+                      throw Error("Error at parsing JSON: %s", buf.ptr());
```
SunFellow, line 307: While try/catch is commented, this exception blocks further attempts to recognize the format.

api/src/indigo_loaders.cpp
SunFellow, line 120: It's hard to identify among a large pull-request that these strings are just moved to better position.
Maybe keep them in-place to improve traceability?

SunFellow (changes requested):
PR needs some cleanup of added, but dead code, and from unnecessary changes before merging.

molecule/src/molecule_auto_loader.cpp
SunFellow: Yeah, treat this as optional proposal.

molecule/molecule_json_loader.h
SunFellow: Yeah, treat this as optional proposal. Can be improved in the future.

api/src/indigo_object.cpp
even1024: fixed

api/src/indigo_savers.cpp
even1024: fixed

api/indigo.h
even1024: fixed

molecule/molecule_json_loader.h
even1024: fixed

molecule/molecule_json_saver.h
even1024: fixed

api/src/indigo_savers.cpp
even1024: fixed

molecule/src/json_loader.cpp
even1024: fixed

molecule/src/molecule_json_loader.cpp
even1024: fixed

api/src/indigo_object.cpp
```diff
@@ -122,8 +122,6 @@ IndigoObjectTypes::IndigoObjectTypes()
     emplace(IndigoObject::TGROUP, "TGroup");
     emplace(IndigoObject::TGROUPS_ITER, "TGroupsIterator");
     emplace(IndigoObject::GROSS_REACTION, "GrossReaction");
-    emplace(IndigoObject::JSON_LOADER, "JsonLoader");
-    emplace(IndigoObject::JSON_SAVER, "JsonSaver");
```
mkviatkovskii: If you are removing it here, also remove corresponding IndigoObject items

SunFellow: https://github.com/epam/Indigo/pull/224#discussion_r571992182

molecule/molecule_json_loader.h
even1024: It has been done intentionally because KET-format had multiple molecules in the previous version and common r-groups list.

molecule/src/molecule_json_loader.cpp
even1024: fixed

molecule/src/molecule_json_loader.cpp
even1024: It doesn't affect functionality. 0 - means no explicitValence.

molecule/src/molecule_json_loader.cpp
even1024: Charge = 0 means no charge. Functionality not affected.

molecule/src/molecule_json_loader.cpp
even1024: The same story: 0 means the element is not an isotope. There is no sense to set it, it's already 0 by default.

molecule/src/molecule_json_loader.cpp
even1024: Checks for allowable values is not implemented for other chemical formats implemented in indigo. Do we really need to do all the checks for KET?

molecule/src/molecule_json_loader.cpp
even1024: It's guaranteed that both _pqmol and _pmol can't be nil.

molecule/src/molecule_auto_loader.cpp
even1024: The format has already been recognised as ket at this point and if the execution will go further without a throwing an exception it fails somewhere in CML parser which is default for all unrecognised formats so it's better to get terminated here.

api/src/indigo_object.cpp
even1024: Fixed

api/src/indigo_loaders.cpp
even1024: fixed

molecule/src/molecule_json_loader.cpp
even1024: fixed

molecule/src/molecule_auto_loader.cpp
even1024: This is workaround was implemented for previous version of ket-format when it had multiple molecule nodes and it was impossible to handle it in one MoleculeJsonLoader class. I'd prefer to leave it as is until we have final ket-format version.

molecule/base_molecule.h
even1024: It came from base_molecule.h where _hl_atoms and _hl_bonds were defined as:

Array<int> _hl_atoms;
Array<int> _hl_bonds;

Moving to std::vector and std::set should be made as a part of another PR related to refactoring and optimisations because there are plenty of places in Indigo where it's possible to see inefficient code like iterating thru arrays instead of using sets.

AlexanderSavelyev: agree that Array<bool> or Array<char> will save some memory. But in terms of performance it will be the same. Do not agree to move to std::vector (and especially set) - as it will decrease a performance. 
In this kind of changes please always consider that number of atoms and bonds is limited and memory optimization has no value.

molecule/src/molecule_auto_loader.cpp
SunFellow: I can't agree that the format has been recognized at this point. How? by finding `{` and then `"molecule"`? It doesn't look as reliable.
If JSON parser would pass syntax parsing, then OK. But this line is exactly throwing on error after `Parse()` method.

even1024: Not a lot of chemical formats has "{" and "molecule" then. So in most cases when we get corrupted KET-file "Error at parsing JSON" will appear which is correct behaviour. There are two options how to make the code better:

1) Remove throw Error("Error at parsing JSON: %s", buf.ptr()) to give a try to interpret the data as some other format and let it fail somewhere else which will tangle the error diagnostic.

2) Implement more reliable KET-detection - it makes sense when we'll have some other chemical data format with "{" and "molecule".

SunFellow: Extracted to separate https://github.com/epam/Indigo/issues/337

PR #227, opened 2020-11-27
Catch Indigo exceptions on `iterate_file`

khyurri:
Now, when IndigoException raises during `iterate_file`, iteration stops with this exception. 

This fix should avoid stop iteration.

What options are:

- Log all errors to stderr
- Accumulate errors and return in another object
- Log error into IndigoRecord 

Open questions:
- Should we save IndigoRecords without fingerprints, when all bits in fingerprint are 0?
- Should we save IndigoRecords without cmf, when IndigoObject cannot be created?

Other fixes:
- Add isort and pylint checks 
- Add check if `poetry update` changes `poetry.lock` file

### DIFF
--- a/.github/workflows/bingo-elastic-ci.yml
+++ b/.github/workflows/bingo-elastic-ci.yml
@@ -13,7 +13,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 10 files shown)


MysterionRise (approved):
Fine with me, please merge

PR #240, opened 2020-12-01
Bugfix/bingo mssql build fix

AlexanderSavelyev:


AlexanderSavelyev:
the fix was deployed and tested

PR #241, opened 2020-12-02
Add context manager and iterator for BingoObject

khyurri:
This PR allows use BingoObject with iterators and context managers.

So this code

```python3
query = indigo.loadQueryMolecule('CC')
sub_matcher = bingo_db.searchSub(query, '')
while sub_matcher.next():
    print(sub_matcher.getCurrentId())
exact_matcher.close()
```

can be rewrited to 

```python3
query = indigo.loadQueryMolecule('CC')
with bingo_db.searchSub(query, '') as sub_matcher:
    for compound in sub_matcher:
        print(compound.getCurrentId())
```

Developer should not worry about closing exact_matcher


MysterionRise (approved):
Looks good to me

PR #261, opened 2020-12-17
api: java: fix load native libs

00Green27:
The name of a resource is a '/'-separated path name that identifies the resource. It means that when specifying the resource file to load, you have to use the forward slash / to separate the path. Since in Windows the value of File.separator is backward slash \, Java won't be able to find the resource file. To make it portable to Windows and some other platforms, you should not use File.separator when loading resource files. Use / instead.

[Class.html#getResourceAsStream(java.lang.String)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getResourceAsStream(java.lang.String))

### DIFF
--- a/api/java/src/main/java/com/epam/indigo/IndigoUtils.java
+++ b/api/java/src/main/java/com/epam/indigo/IndigoUtils.java
@@ -113,18 +113,18 @@ public class IndigoUtils {
     static String getDllPath() {
         String path;
         if (Platform.isWindows()) {
-            path = "Win" + File.separator;
+            path = "Win/";
             if (Platform.is64Bit()) path += "x64";
             else path += "x86";
         } else if (Platform.isMac()) {
-            path = "Mac" + File.separator;
+            path = "Mac/";
             path += "10." + detectMacMinorVersion(path);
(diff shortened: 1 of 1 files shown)


MysterionRise:
@00Green27 unfortunately we have 5 tests failing after those changes

Java's File.separator is https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator

which is `The system-dependent default name-separator character, represented as a string for convenience. This string contains a single character, namely separatorChar.`, so it's safe to use for all platforms

00Green27:
@MysterionRise 
All failed tests failed with an error: **Resource not accessible by integration**. This happens as soon as the person creating the PR does not have permission to the repository.

@MysterionRise File.separator doesn't work for windows. Because the defined path separator in JAR files is /. It is not platform-dependent. You can use / within filenames as well on all platforms in Java. But this isn't a file name, it is a resource name, and the thing it names isn't a file, it is a resource.

mkviatkovskii:
Dear @00Green27,
Thank your for request! I'll test it manually on Windows.

Issue #231, opened 2020-11-30
Result string of indigoData is not properly terminated


achher:
Add data to a DataSGroup:
m.addDataSGroup([0, 1, 2, 3], [], "color", "0.155, 0.55, 0.955")
When reading the data:
nId:= indigoGetDataSGroup(index);
aString:= indigoData(nId)

the resulting string is not terminated properly by 0-character:
'0.155, 0.55, 0.955'#$00AD'º'#$D'ð'#$00AD'º'#$D'ð'#$00AD'º'#$D'ð'#$00AD'º'#$D'ð'#$00AD'ºîþ««««««««îþ'

mkviatkovskii:
Dear @achher,
This issue should be fixed in latest version of Indigo (1.4.3). Could you please take a look and let us know if you are still facing the problem?

achher:
I can confirm the issue is fixed in Indigo (1.4.3). Many thanks.

PR #266, opened 2020-12-18
Fix for issue #231 Result string of indigoData is not properly terminated

SunFellow:
Need to forcibly add terminating zero to `data` Array on reading.
Pros:
1) It is created from zero-terminated C string (`const char *` )
2) Though it was initially designed as Array of bytes and is returned in wrappers as `sbyte *`, it is anyway a string, since no function returns information about its length.
So I conclude it can only be a string, not just an Array of bytes.


SunFellow:
Added python test in todo category reproducing the issue. Unfortunately, the issue is not reproduced before the fix.

PR #267, opened 2020-12-21
Refactoring: Optimize adding atoms and bonds arrays

SunFellow:
- Instead inefficient `Array::push()` one-by-one element, which can cause up to log2(n) reallocations, use `Array::concat`, which reallocates once and then `memcpy` array to array. 
- Some language polishing for `class Array`.

### DIFF
--- a/api/src/indigo_molecule.cpp
+++ b/api/src/indigo_molecule.cpp
@@ -2781,12 +2781,12 @@ CEXPORT int indigoAddDataSGroup(int molecule, int natoms, int* atoms, int nbonds
         int idx = mol.sgroups.addSGroup(SGroup::SG_TYPE_DAT);
         DataSGroup& dsg = (DataSGroup&)mol.sgroups.getSGroup(idx);
         int i;
-        if (atoms != 0)
-            for (i = 0; i < natoms; i++)
-                dsg.atoms.push(atoms[i]);
-        if (bonds != 0)
-            for (i = 0; i < nbonds; i++)
-                dsg.bonds.push(bonds[i]);
+        if (atoms != NULL)
+            dsg.atoms.concat(atoms, natoms);
+
+        if (bonds != NULL)
+            dsg.bonds.concat(bonds, nbonds);
+
         if (data != 0)
             dsg.data.readString(data, true);
         if (name != 0)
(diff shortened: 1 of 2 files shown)


api/src/indigo_molecule.cpp
```diff
@@ -2665,11 +2664,11 @@ CEXPORT int indigoAddSuperatom(int molecule, int natoms, int* atoms,
 int idx = mol.sgroups.addSGroup(SGroup::SG_TYPE_SUP);
 Superatom& satom = (Superatom&)mol.sgroups.getSGroup(idx);
 satom.subscript.appendString(name, true);
-if (atoms == 0)
+if (atoms == NULL)
```
mkviatkovskii, line 2667: Since we are targeting C++11, it's better to use `nullptr` instead.

common/base_cpp/array.h
```diff
@@ -37,19 +37,15 @@ namespace indigo
...
-            _length = 0;
-            _array = NULL;
         }
 
         ~Array()
         {
             if (_array != NULL)
             {
                 free(_array);
-                _array = NULL;
```
mkviatkovskii, line 52: Don't see the reason to delete this line. It may potentially lead to double-free and memory corruption.

SunFellow: Never after the destructor. :)

api/src/indigo_molecule.cpp
SunFellow: Done.

api/src/indigo_molecule.cpp
```diff
@@ -2665,11 +2665,11 @@ CEXPORT int indigoAddSuperatom(int molecule, int natoms, int* atoms,
 int idx = mol.sgroups.addSGroup(SGroup::SG_TYPE_SUP);
 Superatom& satom = (Superatom&)mol.sgroups.getSGroup(idx);
 satom.subscript.appendString(name, true);
-if (atoms == 0)
+if (atoms == nullptr)
     throw IndigoError("indigoAddSuperatom(): atoms were not specified");
 
-for (int i = 0; i < natoms; i++)
-    satom.atoms.push(atoms[i]);
+else
```
mkviatkovskii, line 2671: This `else` is redundant, since other branch of condition leads to exception throwing.

mkviatkovskii:
@SunFellow could you please resolve the merge conflict?

SunFellow:
Resolved conflicts caused by #284

api/src/indigo_molecule.cpp
SunFellow: This is here by two reasons:
1) Replacing `for` it forms in-line comparison of the next line:
```
   satom.atoms.**push**(atoms[i]);	   <----->         satom.atoms.**concat**(atoms, natoms);
```
2) In case the exception is avoided in the future, it is still forms an else clause.

Issue #269, opened 2020-12-22
Update check() method to be consistent with indigo-service check() REST API call


mkviatkovskii:
We have an REST API service written in Python called "indigo-service" (source file is here: https://github.com/epam/Indigo/blob/master/utils/indigo-service/service/v2/indigo_api.py). It has a number of REST API calls, some of them are thin wrappers over Indigo API, some of them contain some extended logic. We want to move all possible logic into C++ part to make it available not only in service.

Currently we have a `check()` method both in service and C++ API (entrypoint is `indigoCheck` in https://github.com/epam/Indigo/blob/master/api/src/indigo_misc.cpp). We need to unite their behavior to provide all required checks to C++ API.

Short example. Indigo API code:
```
from indigo import Indigo
i = Indigo()
qm = i.loadQueryMolecule('''
  Ketcher 12222012222D 1   1.00000     0.00000     0

  6  6  0     0  0            999 V2000
    6.2250   -5.4250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.0910   -5.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.0910   -6.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    6.2250   -7.4250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    5.3590   -6.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    5.3590   -5.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0     0  0
  2  3  2  0     0  0
  3  4  1  0     0  0
  4  5  2  0     0  0
  5  6  1  0     0  0
  6  1  2  0     0  0
M  RBC  1   4   2
M  END
''')
qm.check()
```
It returns empty JSON: `{}`. However, the service call for the same molecule returns
```
{
  "query": "Structure contains query features", 
  "stereo": "Structure has stereochemistry errors"
}
```

So we need to test all features that indigo-service checker affords and improve C++ checker to provide the same behavior.

Note: indigo-service checker could be used through Ketcher (https://lifescience.opensource.epam.com/ketcher/demo.html)

Dmitrii-Chernov-epam-com:
**Steps to reproduce:**

1. Build python api:
	indigo-make-by-libs.py --type=python
	unzip Indigo/dist/indigo-python-1.4.1.5-win.zip to <dir>
[...]

PR #285, opened 2020-12-30
Check function core implementation: #269

Dmitrii-Chernov-epam-com:


api/indigo.h
```diff
@@ -410,6 +410,12 @@ CEXPORT int indigoCheckQuery(int item);
 // Returns one if structure contains RGroup features (RSites, RGroups or attachment points
 CEXPORT int indigoCheckRGroups(int item);
 
+// Returns check result for Indigo object as text file for requested properties as JSON
+CEXPORT const char* indigoCheck2(const char* item, const char* check_flags, const char* load_params);
```
SunFellow, line 414: Better avoid adding new functions those return `const char *` (and gradually kill the old ones) since this pattern provokes to return a pointer to a temporary automatic variable, which lifetime ends with return.

api/tests/c/indigo-test-issue-269.c
```diff
@@ -0,0 +1,139 @@
...
+                                     "    5.3590   -5.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n"
+                                     "  1  2  1  0     0  0\n"
+                                     "  2  3  2  0     0  0\n"
+                                     "  3  4  1  0     0  0\n"
+                                     "  4  5  2  0     0  0\n"
+                                     "  5  6  1  0     0  0\n"
+                                     "  6  1  2  0     0  0\n"
+                                     "M  RBC  1   4   2\n"
+                                     "M  END\n";
+
```
SunFellow, line 26: How about to switch to CPP file and utilize raw string literals (available since C++11)?
Then the test data would look closer to the original file, like:
```
const char* FAILED_MOLECULE_STRING = R"(The molecule
  Ketcher 12222012222D 1   1.00000     0.00000     0

  6  6  0     0  0            999 V2000
    6.2250   -5.4250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.0910   -5.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.0910   -6.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    6.2250   -7.4250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    5.3590   -6.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    5.3590   -5.9250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0     0  0
  2  3  2  0     0  0
  3  4  1  0     0  0
  4  5  2  0     0  0
  5  6  1  0     0  0
  6  1  2  0     0  0
M  RBC  1   4   2
M  END
)";

```

molecule/structure_checker2.h
```diff
@@ -0,0 +1,150 @@
...
+#include <unordered_set>
+
+class IndigoObject;
+
+namespace indigo
+{
+
+    class BaseMolecule;
+
+    class DLLEXPORT StructureChecker2
```
SunFellow, line 35: Is this class to supersede `StructureChecker`? Otherwise, please do not repeat nearly the same enums, moreover public, but reuse them (e.g. by inheritance or extracting to some common header...)

molecule/structure_checker2.h
```diff
@@ -0,0 +1,150 @@
...
+            CHECK_3D_COORD = 0x00002000,     // Check 3D coordinates existance
+            CHECK_CHARGE = 0x00004000,       // Check charged structure
+            CHECK_SALT = 0x00008000,         // Check possible salt structure
+            CHECK_AMBIGUOUS_H = 0x00010000,  // Check ambiguous H existance
+            CHECK_COORD = 0x00020000,        // Check coordinates existance
+            CHECK_V3000 = 0x00040000,        // Check v3000 format
+            CHECK_ALL = -1                   // Check all features (default)
+        };
+
+        static constexpr const char* checkTypeName[] = {"none",         "load",         "valence", "radical",    "pseudoatom", "stereo",    "query",
```
SunFellow, line 63: Better use a map from the beginning.

SunFellow: In addition, 
* if these literals are subject of interface, please expose them as named strings (or macros);
* if these literals are internal matter, please hide the initialization to cpp file.

api/src/indigo_internal.h
```diff
@@ -162,6 +162,7 @@ class DLLEXPORT IndigoObject
         INDIGO_OBJECT_LAST_TYPE // must be the last element in the enum
     };
 
+    int id;
```
SunFellow, line 165: This change looks attractive, but in fact is discussible. (For example, what would you do, if the stored id mismatches with id key ).
@mkviatkovskii , what do you think?

api/src/indigo_molecule.h
```diff
@@ -47,7 +47,7 @@ class DLLEXPORT IndigoBaseMolecule : public IndigoObject
 
     const char* debugInfo();
 
-    static bool is(IndigoObject& object);
+    static bool is(const IndigoObject& object);
```
SunFellow, line 50: There are some changes, useful, but not related to the subject of the pull-request/task.
It's better to extract them in a separate pr.

api/src/indigo_molecule.cpp
```diff
@@ -54,7 +54,7 @@ IndigoMolecule::IndigoMolecule() : IndigoBaseMolecule(MOLECULE)
 {
 }
 
-bool IndigoBaseMolecule::is(IndigoObject& object)
+bool IndigoBaseMolecule::is(const IndigoObject& object)
```
SunFellow, line 57: There are some changes, useful, but not related to the subject of the pull-request/task.
It's better to extract them in a separate pr.

molecule/src/molecule_name_parser.cpp
```diff
@@ -531,8 +531,18 @@ void MoleculeNameParser::FragmentNodeBase::print(ostream& out) const
     out << "Type: FragmentNodeBase" << endl;
 
     out << "Multipliers:" << endl;
-    const auto& mul_container = multipliers._Get_container();
-    std::for_each(mul_container.begin(), mul_container.end(), [&out](const Multiplier& multiplier) { out << "\tvalue: " << multiplier.first << endl; });
+    auto muls = multipliers;
+    Multipliers muls_rev;
+    while (!muls.empty())
+    {
+        muls_rev.push(muls.top());
```
SunFellow, line 538: Isn't possible to use reverse iteration?

molecule/src/molecule_name_parser.cpp
```diff
@@ -531,8 +531,18 @@ void MoleculeNameParser::FragmentNodeBase::print(ostream& out) const
...
+        muls.pop();
+    }
+    while (!muls_rev.empty())
+    {
+        out << "\tvalue: " << muls_rev.top().first << endl;
+        muls_rev.pop();
+    }
```
SunFellow, line 545: Maybe better change the container type of Multipliers, than transfuse the stack twice, not counting the initial copy-in-assignment.

SunFellow (commented):
Generic considerations:
- Please give distinct title to the pull request and to commits' messages. (Sometimes it's preferable to watch the differences commit-by-commit, but empty messages don't give a clue to what the commit is about)
- There are some changes, useful in general, but not related to the subject of the pull-request/task. It's better to extract such ones in a separate pr.
- Unclear relation between the new checks #2 and the previous checks. Are the new ones to replace the old once? or to extend the functionality? There are concerns about enums repeated in the new class. Probably, some other elements could be confusing either.

api/tests/c/indigo-test-issue-269.c
Dmitrii-Chernov-epam-com: This test emulates the check() call from C API code for host lang (Java, Python etc.). So its  compilability as C is a subject of testing. I 'm not sure that this minor cosmetics is enough to skip it.

molecule/structure_checker2.h
Dmitrii-Chernov-epam-com: Yes, the StructureChecker2.* is full replacement for StructureChecker.* and ReactionChecker.*

The functions "check2" in C and Python replace the corresponding functions "check"

After debugging, "check" instances should be removed and "check2" instances should be renamed to "check"

molecule/structure_checker2.h
Dmitrii-Chernov-epam-com: Mostly, these literals are here for self-documenting: they should be used in host language (J, Py etc.) and never in C/C++ code. C/C++ should use enum constants. If consts or macroses look clearer for these strings, OK, I'll rewrite it.

In the current code this is unclear thing: Python uses strings but which stings are available and how they map to the enum can be investigated by reverse engineering only.

Using of map is possible but redundant here (see **check_params_from_string(const char* params)**).

api/src/indigo_internal.h
Dmitrii-Chernov-epam-com: The id is assigned here and only here on object creation. If it changed during the object lifecycle, it would be a bug in many places because the id identifies object for host language (Java.Py etc.) code. So, it can't mismatch or nothing works.

I've introduced the field to walk around architectural inconsistency:  in C++ API, Indigo objects sometimes are referenced by handle (i.e. the id), sometimes by pointer.

int handle => IndigoObject pointer conversion is provided by indigoGetInstance().getObject(int item_handle)
but
IndigoObject pointer => int handle conversion is not provided.

Particularly for me, the problem is here:
static void checkMolecule(**const IndigoObject& item**, int check_types, const std::vector<int>& selected_atoms, const std::vector<int>& selected_bonds,
                          StructureChecker2::CheckResult& result)
{
    checkMolecule(((IndigoObject&)item).getBaseMolecule(), check_types, selected_atoms, selected_bonds, result);
    **check_v3000(item._id,_ result);**
}

static void check_v3000(int mol, StructureChecker2::CheckResult& result)
{
    const char* f = **indigoMolfile(mol);** _<= handle required, no API for IndigoObject&_
    if (f && std::string(f).find("V3000") != -1)
......
}

api/src/indigo_molecule.h
Dmitrii-Chernov-epam-com: The changes are related to the PR due to it requires them to avoid compile errors.
Other code is not affected.
But if you don't want them, I can rollback them and do workaround.

api/src/indigo_molecule.cpp
Dmitrii-Chernov-epam-com: The changes are related to the PR due to it requires them to avoid compile errors.
Other code is not affected.
But if you don't want them, I can rollback them and do workaround.

molecule/src/molecule_name_parser.cpp
Dmitrii-Chernov-epam-com: Maybe better but refactoring of molecule_name_parser.cpp is a separate task.

I've patched compilation error for VS2019 (const auto& mul_container = multipliers._Get_container();<=_Get_container is deprecated) with minimal code intervention. If you want something more, please create an issue and assign it to me.

Dmitrii-Chernov-epam-com:
The class StructureChecker2
replaces StructureChecker and ReactionChecker classes.

The functions "check2" in C and Python replace the corresponding functions "check"

After debugging, "check" instances should be removed and "check2" instances should be renamed to "check"

api/tests/c/indigo-test-issue-269.c
SunFellow: OK, agree.

molecule/structure_checker2.h
SunFellow: OK. So finally (merged or squashed), we will be able to trace differences in-place (in structure_checker.h, in structure_checker.cpp, etc.)?

Dmitrii-Chernov-epam-com: Yes but this difference is up to 100%.
I've re-organized this "spaghetti" code so that each check is a separate independent (and unit-testable) function. See the functions referred in this array: 
static void (*check_type_checkers[])(Molecule&, const std::unordered_set<int>&, const std::unordered_set<int>&, StructureChecker2::CheckResult&) = {
    &check_load,         &check_valence,      &check_radical, &check_pseudoatom, &check_stereo,     &check_query,
    &check_overlap_atom, &check_overlap_bond, &check_rgroup,  &check_sgroup,     &check_tgroup,     &check_chirality,
    &check_chiral_flag,  &check_3d_coord,     &check_charge,  &check_salt,       &check_ambigous_h, &check_coord};

And joined structure_checker and reaction_checker with automatic detection of kind of checked item.

So the code is taken from the current checkers and python REST code but its fragments are combined by another way and git can't find them in prev version.

Dmitrii-Chernov-epam-com: >please do not break the history of the original file(s), i.e. do not use git rm
OK, I'll just copy-paste structure_checker2.* content to structure_checker.* and delete structure_checker2.* and reaction_checker.*

But this should be done after final approve only as a final cleanup. Now we can see, run and compare the both codes and this is helpful for testing & debugging.

molecule/structure_checker2.h
SunFellow: Can Python/Java catch the strings from C interface file as well as from DLL interface? Probably, not. In this case, it's unnecessary to expose the strings, but enough to mention them in the documentation comment.

Dmitrii-Chernov-epam-com: > mention them in the documentation comment.

OK, will do.

Dmitrii-Chernov-epam-com: But this makes possible an inconsistency between doc comment and code if one of them is changed and another one forgotten to update.

api/src/indigo_internal.h
SunFellow: Yep, you wrote exactly what I meant by "looks attractive".
When I talked about mismatch, I meant mismatch at the higher level: between input parameter `id` in calls to functions those search by `int id` and found object's filed `id`. If I'm not mistaken, this input parameter is a key in internal maps, so it exists separately. Finally, having `id` in two places is a redundancy (or indexing).

api/src/indigo_molecule.h
SunFellow: OK. No problem.

molecule/src/molecule_name_parser.cpp
SunFellow: OK. will create an issue for optimization. Closing this.

api/src/indigo_molecule.cpp
SunFellow: OK. No problem.

molecule/structure_checker2.h
SunFellow: Yes, this is possible in general with any documentation. But what are alternatives?
We could provide a function that returns a list of recognizable strings of check types, but the caller anyway needs to know the semantics from some documentation. It can only use such a list for additional check before call.

api/src/indigo_internal.h
Dmitrii-Chernov-epam-com: Yes, id is the object's unique permanent key in internal map.
Now object knows its id.
That's all.
What "mismatch" is possible?

int Indigo::addObject(IndigoObject* obj)
{
    OsLocker lock(_objects_lock);
    obj->id = _next_id++;
    _objects.insert(obj->id, obj);
    return obj->id;
}

molecule/src/molecule_name_parser.cpp
SunFellow: Oh, just realized that this change is a fix for `_Get_container` failed compilation in latest Visual Studio and that was hidden under `#ifdef DEBUG` . So it didn't harm release builds.
I hope this change is quite isolated from other changes in this pull-request, right?
How about the following: make a separate pull-request for this fix alone and also try to remove `#ifdef DEBUG`? So the class (`FragmentNodeSubstituent`) would be the same for all platforms and configurations.

api/tests/c/indigo-test-issue-269.c
```diff
@@ -0,0 +1,139 @@
...
+                                   " 29 30  1  0     0  0\n"
+                                   " 15 31  1  1     0  0\n"
+                                   " 29 32  1  1     0  0\n"
+                                   " 28 33  1  1     0  0\n"
+                                   " 16 34  1  1     0  0\n"
+                                   "  6 35  1  1     0  0\n"
+                                   " 11 36  1  1     0  0\n"
+                                   "M END\n";
+
+int test(int negative)
```
SunFellow, line 104: Nevertheless, instead parametrization for test scenarios, it's better to use different functions with distinctive names. So, instead `test(int negative)`, I'd make two tests like (just a proposal): 
`testLoadGoodMoleculeSucceeds()`
`testLoadBadMoleculeFails()`
without a parameter and without branching the flow.
Yes, this increases duplicate code index, but it's inherent for tests.

api/tests/c/indigo-test-issue-269.c
```diff
@@ -0,0 +1,139 @@
...
+    {
+        fprintf(stderr, "Error: can't parse input; %d is returned by indigoLoadQueryMoleculeFromString\n", qm);
+    }
+    const char* result = indigoCheck(qm, 0);
+    printf(!result ? "NULL" : result);
+    indigoFree(qm);
+    return 0;
+}
+
+int test2(int negative)
```
SunFellow, line 117: Same.

molecule/structure_checker2.h
Dmitrii-Chernov-epam-com: >But what are alternatives?
Please, look the alternative commited: public text constants and private string array. Good style and they can't be changed separately. C++ programmers typically shouldn't use the constants but they can.

api/tests/c/indigo-test-issue-269.c
Dmitrii-Chernov-epam-com: Please, don't review this file now.
This is not tests yet (I don't have data for them), this is just a dummy startup code for debugging of my code.
It will be rewritten

molecule/src/molecule_name_parser.cpp
Dmitrii-Chernov-epam-com: The separate PR: https://github.com/epam/Indigo/pull/294

This is a debugging output only, don't worry. Prod is not affected.

molecule/structure_checker2.h
SunFellow: Just noticed that original comments contained spelling errors. They are copied as is to the new comments.
Could you please check and fix spelling once you are changing this interface files?
There are mistypes: existance, fetaures, strerochemistry.

Dmitrii-Chernov-epam-com: Fixed, pushed

molecule/src/structure_checker2.cpp
```diff
@@ -0,0 +1,651 @@
...
+                        StructureChecker2::CheckResult& result)
+{
+    if (mol.vertexCount() > 1 && !BaseMolecule::hasCoord(mol))
+    {
+        result.message(StructureChecker2::CheckMessageCode::CHECK_MSG_ZERO_COORD);
+    }
+}
+static void check_v3000(int mol, StructureChecker2::CheckResult& result)
+{
+    const char* f = indigoMolfile(mol);
```
mkviatkovskii, line 548: there is a dependency violation here - core libraries (common, graph, molecule, layout, reaction) should not depend on API. This works for you now only because you don't build Bingo (which depends on molecule, but not on API).

And also ID only makes sense in Indigo session, not even in IndigoObject - objects could potentially be even serialized/deserialized between sessions and they should not now their ID.

So I suggest passsing here `BaseMolecule` instead of id.

molecule/src/structure_checker2.cpp
```diff
@@ -0,0 +1,651 @@
...
+                                           "Input structure is empty",
+                                           "Structure contains ambiguous hydrogens",
+                                           "Structure contains query features, so ambiguous H could not be checked",
+                                           "Structure contains 3D coordinates",
+                                           "Structure has no atoms coordinates"
+                                           "Reaction component check result",
+                                           "Not implemented yet: check chirality",
+                                           "Not implemented yet: check charge",
+                                           "Not implemented yet: check salt",
+                                           "Structure supports only Molfile V3000"
```
mkviatkovskii, line 644: Not sure why do we need this array - we can instead write these string directly in methods where we address them via enums. Do you think code looks more clear in this approach?

molecule/src/base_molecule.cpp
```diff
@@ -58,6 +58,35 @@ QueryMolecule& BaseMolecule::asQueryMolecule()
 
 bool BaseMolecule::isQueryMolecule()
 {
+    bool r = false;
+    for (auto idx : vertices())
```
mkviatkovskii, line 62: `const auto`

molecule/src/base_molecule.cpp
```diff
@@ -58,6 +58,35 @@ QueryMolecule& BaseMolecule::asQueryMolecule()
...
+    {
+        r = r || (reaction_atom_exact_change[idx] || reaction_atom_inversion[idx]);
+    }
+
+    for (auto idx : edges())
```
mkviatkovskii, line 67: const auto

molecule/src/structure_checker2.cpp
```diff
@@ -0,0 +1,651 @@
...
+StructureChecker2::CheckResult StructureChecker2::check(const char* item, const char* check_flags, const char* load_params)
+{
+    int it = indigoLoadStructureFromString(item, load_params);
+    if (it < 0)
+    {
+        it = indigoLoadStructureFromString(item, (std::string(load_params ? load_params : "") + " query").c_str()); //##!!!PATCH
+    }
+    auto r = check(it, check_flags);
+    indigoFree(it);
+    return r;
```
mkviatkovskii, line 114: Same here as in below comment - we should not use Indigo API in core libraries. We need to load molecule from string in API and pass `BaseMolecule` here instead of string.

molecule/src/structure_checker2.cpp
```diff
@@ -0,0 +1,651 @@
...
+            while (std::regex_search(a, sm3, rx3))
+            {
+                vec.push_back(atoi(sm3[1].str().c_str()));
+                a = sm3.suffix();
+            }
+            s = sm2.suffix();
+        }
+    }
+    return r;
+}
```
mkviatkovskii, line 93: This logic is pretty complicated, and i think that for first iteration we should not parse this params, and `check2` should just perform all checks for all atoms and bonds.
Atoms and bonds selection will be later done via `selected` field in BaseMolecule (WIP in another branch).

molecule/src/base_molecule.cpp
Dmitrii-Chernov-epam-com: OK

molecule/src/base_molecule.cpp
Dmitrii-Chernov-epam-com: OK

molecule/src/structure_checker2.cpp
Dmitrii-Chernov-epam-com: Regarding to dependency violations:
image
"Check" functionality is spreaded across the code.
molfile_loader performs some checks itself instead of loading of any syntactically correct file assuming that the result may/should be checked later by the "checker" which concentrates all checking functionality.

Therefore you can't check molecule containing errors, i.e. exactly for what checking is necessary: you just can't load it without not-evident dancing around "ignore errors" options.

Also note that "ex" is added by me:
Originally it throws anonymous exception and fails silently:
       catch (Error&)
        {
            if (!options.ignore_errors)
                throw;
        }

Dmitrii-Chernov-epam-com: Regarding to dependency violation:

I see that API implements features - see below.
Core should call API for the feature? - this is a dependency violation
API is not an interface to expose core to users but it implements features itself? How to use the features in core programming? Duplicate them?

In the particular case below, why we write the "checker" if we don't use it for checking?
Checking facilities appear in many unpredictable places (see above and below) including Python REST service. 


CEXPORT int indigoCheck3DStereo(int item)
{
    INDIGO_BEGIN
    {
        IndigoObject& obj = self.getObject(item);

        if (IndigoBaseMolecule::is(obj))
        {
            BaseMolecule& bmol = obj.getBaseMolecule();

            bool stereo_3d = true;

            if ((bmol.stereocenters.size() > 0) && BaseMolecule::hasZCoord(bmol))
            {
                for (auto i : bmol.vertices())
                {
                    if (!bmol.stereocenters.exists(i))
                    {
                        const Vertex& vertex = bmol.getVertex(i);

                        for (auto j = vertex.neiBegin(); j != vertex.neiEnd(); j = vertex.neiNext(j))
                            if (bmol.getBondDirection2(i, vertex.neiVertex(j)) > 0)
                                stereo_3d = false;
                    }
                }
            }
            else
                stereo_3d = false;

            if (stereo_3d)
                return 1;
        }
        return 0;
    }
    INDIGO_END(-1);
}

Dmitrii-Chernov-epam-com: Indigo/api/src/indigo_stereo.cpp

500 rows of code. 
Including private funcs.
This is not API, this is core functionality

Should core use it? 
Should core call it using handles?
CEXPORT int indigoValidateChirality(**int handle**)

molecule/src/structure_checker2.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+bool StructureChecker2::CheckResult::isEmpty() const
+{
+    return messages.size() == 0;
+}
+
+static void message(StructureChecker2::CheckResult& result, StructureChecker2::CheckMessageCode code, int index, std::vector<int> ids,
+                    const StructureChecker2::CheckResult& subresult)
+{
+    StructureChecker2::CheckMessage m = {code, index, ids, subresult};
```
SunFellow, line 45: Unfortunately, 2 of 3 Indigo CI checks fail at this line.

### build_mac/Build native libs:
```
/Users/runner/work/Indigo/Indigo/molecule/src/structure_checker2.cpp:45:37: error: no matching constructor for initialization of 'StructureChecker2::CheckMessage'
     StructureChecker2::CheckMessage m = {code, index, ids, subresult};
                                     ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/Indigo/Indigo/molecule/../molecule/structure_checker2.h:109:26: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 4 were provided
```
### build_linux/Build native libs:
```
/__w/Indigo/Indigo/molecule/src/structure_checker2.cpp:45:69: error: could not convert '{code, index, ids, subresult}' from '<brace-enclosed initializer list>' to 'indigo::StructureChecker2::CheckMessage'
   45 |     StructureChecker2::CheckMessage m = {code, index, ids, subresult};
      |                                                                     ^
      |                                                                     |
      |                                                                     <brace-enclosed initializer list>
```
Possible reason can be due to inline initialization in the class: 
```
        struct DLLEXPORT CheckMessage
        {
            CheckMessageCode code = StructureChecker2::CheckMessageCode::CHECK_MSG_NONE;
...
            int index = -1;
```

See discussion: https://stackoverflow.com/questions/37776823/could-not-convert-from-brace-enclosed-initializer-list-to-struct

Dmitrii-Chernov-epam-com: Fixed

SunFellow (approved):

api/src/indigo_structure_checker.h
```diff
@@ -0,0 +1,86 @@
...
+    public:
+        /// <summary>
+        /// Textual Check Type values for the corresponding CheckTypeCode
+        /// to be used in host language calls (Java, Python etc.) as a comma-separated list
+        /// See below check(char *, char *):
+        /// check("molecule text", "load, valence, radical, atoms 1 2 3, bonds 4 5 6, tgroup"
+        /// is equivalent to
+        /// check(the_molecule, check_flags = CHECK_LOAD | CHECK_VALECE | CHECK_RADICAL | CHECK_TGROUP, selected_atoms=[1,2,3], selected_bonds=[4,5,6]);
+        /// </summary>
+        static constexpr char* CHECK_NONE_TXT = "none";                 // Check none
```
AlexanderSavelyev, line 46: Code duplicate for the check types. It is better to keep all the check types in structure_checker2. Then remove all this strings

api/src/indigo_structure_checker.cpp
```diff
@@ -0,0 +1,205 @@
...
+using namespace indigo;
+
+struct CheckParams
+{
+    int check_flags = StructureChecker2::CHECK_ALL;
+    std::vector<int> selected_atoms;
+    std::vector<int> selected_bonds;
+};
+
+static constexpr const char* checkTypeName[] = {
```
AlexanderSavelyev, line 40: code duplicate. It is worth to keep all the check types in structure_checker2 (see comment below)

api/src/indigo_structure_checker.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+            words.insert(sm1[1]);
+            s = sm1.suffix();
+        }
+        r.check_flags = StructureChecker2::CHECK_NONE;
+        for (int i = 0; i < len; i++)
+        {
+            if (words.find(std::string(checkTypeName[i])) != words.end())
+            {
+                r.check_flags |= 1 << i;
```
AlexanderSavelyev, line 69: Overcomplicate logic. In order to add a new type to check one need to go to 4 places and append correct bitwise value. Instead, it is suggested to have mapping between codes and strings (see comment below)

api/src/indigo_structure_checker.cpp
```diff
@@ -0,0 +1,205 @@
...
+    IndigoStructureChecker::CHECK_RADICAL_TXT,   IndigoStructureChecker::CHECK_PSEUDOATOM_TXT,   IndigoStructureChecker::CHECK_STEREO_TXT,
+    IndigoStructureChecker::CHECK_QUERY_TXT,     IndigoStructureChecker::CHECK_OVERLAP_ATOM_TXT, IndigoStructureChecker::CHECK_OVERLAP_BOND_TXT,
+    IndigoStructureChecker::CHECK_RGROUP_TXT,    IndigoStructureChecker::CHECK_SGROUP_TXT,       IndigoStructureChecker::CHECK_TGROUP_TXT,
+    IndigoStructureChecker::CHECK_CHIRALITY_TXT, IndigoStructureChecker::CHECK_CHIRAL_FLAG_TXT,  IndigoStructureChecker::CHECK_3D_COORD_TXT,
+    IndigoStructureChecker::CHECK_CHARGE_TXT,    IndigoStructureChecker::CHECK_SALT_TXT,         IndigoStructureChecker::CHECK_AMBIGUOUS_H_TXT,
+    IndigoStructureChecker::CHECK_COORD_TXT,     IndigoStructureChecker::CHECK_V3000_TXT,        IndigoStructureChecker::CHECK_ALL_TXT};
+
+static CheckParams check_params_from_string(const char* params)
+{
+    CheckParams r;
```
AlexanderSavelyev, line 51: It is worth to avoid one letter variable names

molecule/src/structure_checker2.cpp
AlexanderSavelyev: this array has specific order for the CheckMessageCode which can be difficult to support. Instead it is worth replace the message_list to message_map<CheckMessageCode , string>

molecule/structure_checker2.h
```diff
@@ -0,0 +1,133 @@
...
+
+    class BaseMolecule;
+    class BaseReaction;
+
+    class DLLEXPORT StructureChecker2
+    {
+    public:
+        enum CheckTypeCode
+        {
+            CHECK_NONE = 0x00000000,         // Check none
```
AlexanderSavelyev, line 39: duplicate code and overcomplicate logic. Instead it is worth to create two maps with 
types_map<string, CheckMessageCode>
messages_map<CheckMessageCode , string>

api/src/indigo_structure_checker.h
AlexanderSavelyev: And use types_map<CheckMessageCode , string>

api/tests/c/indigo-test-data-issue-269.cpp
```diff
@@ -0,0 +1,1061 @@
+extern "C" const char* issue269test_RGroupsWarning = R"(
```
AlexanderSavelyev, line 1: There are warnings in g++ for this file. It is worth to move the test into the python tests

api/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ if (NOT PACK_INDIGO_NOT)
 endif()
 if (NOT DEFINED ENV{DISABLE_INDIGO_TESTS})
 DEFINE_TEST(indigo-c-test-shared "tests/c/indigo-test.c" indigo-shared)
+DEFINE_TEST(indigo-c-test-shared-issue-269 "tests/c/indigo-test-issue-269.c;tests/c/indigo-test-data-issue-269.cpp" indigo-shared)
```
AlexanderSavelyev, line 71: this tests are related to indigo c API. The tests for checker need to be moved to python tests

AlexanderSavelyev:
3 tests fails from api/tests/python @Dmitrii-Chernov-epam-com  need to fix python tests first

api/CMakeLists.txt
Dmitrii-Chernov-epam-com: OK but copied.
If something went wrong, these C tests would be a helpful entry point for C++ debugging.

AlexanderSavelyev: I would suggest to restore cpp unit tests. I will create additional PR for that

Dmitrii-Chernov-epam-com: Copied to Python. 
Done.

api/src/indigo_structure_checker.cpp
Dmitrii-Chernov-epam-com: This is not code duplicate.
structure_checker2 works with bitwise values only and knows nothing about text values. 
text values are used by indigo_structure_checker only; structure_checker2 doesn't parse them 

So, it's not logical to define text values in structure_checker2 - they are related to API level and are not known in Core.

api/src/indigo_structure_checker.cpp
Dmitrii-Chernov-epam-com: When I worked in embedded software development (see MStar inc. in my CV), it was a strict coding style requirement to have exactly one return statement at the end of function.
E.g. let you have to add a finalization code, free resources etc. You add the code at the end of func..... oops! it has 100500 returns from its body and you even can't be sure that you've found all of them.

I am used to following this style.
In my code "var r" always means return value like "var i" means iterator.

I write

bool r;
if (smth)
  r = true;
else
 r = false
return r;

instead of

if (smth)
  return true;
else
 return false

Of course, it's possible to change "r" to e.g. "ret". But it appears in many places and this requires significant work.

api/tests/python/tests/check/check.py
```diff
@@ -0,0 +1,41 @@
...
+    { "name" : "issue269test_Issue_293_All",  "test": "radical;pseudoatom;stereo;query;overlap_atom;overlap_bond;rgroup;chirality;3d_coord", "expected": '''[{"code":6,"message":"Structure contains radicals","ids":[13,15,17]},{"code":14,"message":"Structure contains query features"}]'''},
+    ] 
+
+errors = ''
+for test in tests:
+    print(f"\nTEST: {test['name']}\n")
+    with open(joinPath(f"molecules/{test['name']}.mol"), 'r') as file:
+        molfile = file.read()
+    res = indigo.check2(molfile, test['test'])
+    if (res == test['expected']):
```
SunFellow, line 26: You can follow the (quite interesting) approach used here in api python tests. 
Instead defining expected data and comparing with results inside the test, each test has its reference expected output (ref/**/test*.out) and the frame above compares the output produced by each test with this expected reference.

AlexanderSavelyev (approved):
PR was reviewed. Later it is worth to rename the check2 to check ( deprecate old one) and move the c tests into the unit tests. But it can be done by separate PR

PR #287, opened 2020-12-30
Fix `Array::memcmp`

SunFellow:
Fix `Array::memcmp` according to VCRT::memcmp: the content should prevail the length

### DIFF
--- a/common/base_cpp/array.h
+++ b/common/base_cpp/array.h
@@ -183,15 +183,15 @@ namespace indigo
 
         int memcmp(const Array<T>& other) const
         {
+            size_t min_length = std::min(_length, other._length);
+            int result = ::memcmp(_array, other._array, min_length * sizeof(T));
+            if (result != 0)
+                return result;
             if (_length < other._length)
                 return -1;
             if (_length > other._length)
-                return -1;
-
-            if (_length == 0)
-                return 0;
-
-            return ::memcmp(_array, other._array, _length * sizeof(T));
+                return +1;
+            return 0;
         }
 
         void remove(int idx, int span = 1)
--- a/common/base_cpp/array.h
+++ b/common/base_cpp/array.h
@@ -183,7 +183,7 @@ namespace indigo
 
         int memcmp(const Array<T>& other) const
         {
-            size_t min_length = std::min(_length, other._length);
+            size_t min_length = __min(_length, other._length);
             int result = ::memcmp(_array, other._array, min_length * sizeof(T));
             if (result != 0)
                 return result;


mkviatkovskii (commented):
I'm not sure that this is required. Indigo Array is not just a c-string that should be first of all compared in a lexicographic order. 
And here we are attempting to change method's behavior without any specific reason for that. For instance, we have two `Array<int>`:
```
{0, 1, 2}
{0, 1, 1, 1}
```
Current implementation returns `-1`, since first array length is lesser.
Your implementation will return `+1` since first different byte in first array is greater.

This could break the logic of algorithms that use `Array::memcmp`, and there are about 6 usages in Indigo code.
If you insist on this changes, we should explicitly test all usage cases and check that algorithms still work in the same way and we haven't broken anything.

SunFellow:
If the function has different semantics, it should have different name. Something like `less`, however, I'm not sure that it also compare the length first... How about to rename it?

Fortunately or not, all usages of Array::memcmp() are now compared against 0.
The two cases when it is compared to -1 are now commented. 

We need to ask algorithm author (@IuriiPuzanov?) what was it used for?

AlexanderSavelyev:
@SunFellow it seems that original function verifies the length of this and external buffer and then if the length is equal calls the memcmp, overwise returns -1. For me it seems a correct behavior (e.g. if lengths are different, then it can be memory corruption).
Could you point the reason for the suggested change. It seems like copy and shrink to min length can corrupt data. It makes sense to have something like clone() ( by resize and copy) but why do you need to change this method then. Please advise for the change reason, because it is not clear?

SunFellow:
@AlexanderSavelyev , the reason is simple: to make Array::memcmp() comply with RTL or std:: memcmp() function.
If you see the familiar name, you expect / assume the same behavior.

memcmp specifications say:

**C memcmp:**

> Compares the first count characters of the objects pointed to by lhs and rhs. The comparison is done lexicographically.
> 
> The sign of the result is the sign of the difference between the values of the first pair of bytes (both interpreted as unsigned char) that differ in the objects being compared.
> 

**C++ std::memcmp:**

> Reinterprets the objects pointed to by lhs and rhs as arrays of unsigned char and compares the first count characters of these arrays. The comparison is done lexicographically.
> 
> The sign of the result is the sign of the difference between the values of the first pair of bytes (both interpreted as unsigned char) that differ in the objects being compared.
> 

It's pity, but there is nowhere in the world said what the function does, if lhs and rhs point to buffers of different sizes. So we can only guess that if we pass zero-terminated strings, then the longer string will be treated greater, since any byte is greater than 00h.
For cases like Mikhail's sample, yes, the behavior changes. 
And if put the sample to the library memcp(), it will also return different result: the first mismatch byte is at 2 position: 2 > 1, then return > 0.

mkviatkovskii:
@SunFellow `std::memcmp` is defined in `<cstring>` and operates on c-strings (arrays of `unsigned char`). And according to documentation "The comparison is done lexicographically". Interpreting arrays of, for instance, `double` as arrays of `unsigned char` and their lexicographical comparison doesn't make any sense.

As soon as our `Array` is much more than just array of `unsigned char`, I don't think we should merge this request.

SunFellow:
I don't understand your point as a reason to drop the request. The difference is not in interpreting lexicographical comparison - here nothing changes, the difference is how to handle the different length/size of arrays being compared.

PR #294, opened 2021-01-14
_Get_container() deprecation fix

Dmitrii-Chernov-epam-com:
_Get_container() is deprecated in VS-2019.
This PR workarounds it to avoid compile error.

The issue is inside \#ifdef DEBUG section, this is a debugging output only.
Therefore its performance and effectiveness are not a matter.

**No production code is affected**.


SunFellow:
Now Debug configuration compiles! :)

SunFellow (approved):
Now Debug (VS 2019) configuration compiles! :)

PR #300, opened 2021-01-20
Fix warnings about `double` to `float` conversion: MSVC C4244, C4305.

SunFellow:
MSVC compilation produces a lot of warnings. Significant amount is dedicated to floating-point conversion between double precision (which is default) to single precision. 
This fix aims decreasing amount of such warnings:

- by replacing implicit conversion to explicit conversion without losing the sense and precision in expressions;
- when assigning or initializing float variables;
- when passing function float parameters;
- by changing literals from double precision or casted to double precision (by default) to single precision literals (like `1.f`) where it keeps calculation in single precision.
- kept double at some places, where precision can be critical, for example in circles, and for constant pi.

For explicit conversion, introduced macro `_2FLOAT(x)`, which resolves to `static_cast<float>(x)` operator under C++ or to C-style type conversion `(float)(x)` under C. 
The resulting numbers of these specific warnings are as follows:

Warning | Before fix | After fix 
--------- | ---------- | ----------
C4305 (truncation) | 255 | **0**
C4244 (conversion, possible loss of data) | 670 | **144**

The resting warnings C4244 mostly refer to integral types conversion and 64-bit pointers and `size_t` problems that requires a separate issue.
Also, 3-d party code under `./third_party/` path is not touched.


AlexanderSavelyev (approved):
In general no issues with the PR. The only suggestion is to include the warnings check for other compilers (not only VS)

PR #315, opened 2021-01-31
_Get_container() deprecation fix in Release as well

SunFellow:
Removed `#ifdef DEBUG` conditional compilation around `print()` methods.
Now the code does not diverge depending on configuration.
However, print() methods are not used, so this won't affect Release configuration.
This pull request is on top of #294.

### DIFF
--- a/molecule/molecule_name_parser.h
+++ b/molecule/molecule_name_parser.h
@@ -48,9 +48,7 @@
 
 
 
-#ifdef DEBUG
 #include <iostream>
-#endif
 
 #include "base_cpp/non_copyable.h"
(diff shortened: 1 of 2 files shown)


mkviatkovskii:
Don't see the reason to remove `#ifdef DEBUG`. Since these are debug methods, we don't need to have them in release build.

SunFellow:
> do you have any specific notes on this?

Nothing but mentioned in the description:

> Now the code does not diverge depending on configuration.

mkviatkovskii:
OK, then I'm closing the request, since these are debug-only methods.

PR #329, opened 2021-02-16
api: dotnet: expose oneBitsList

f1nzer:
oneBitsList was not mapped in dotnet

related java issue: https://github.com/epam/Indigo/issues/191

ps: also added necessary gitignore patterns for dotnet tools

### DIFF
--- a/api/dotnet/IndigoObject.cs
+++ b/api/dotnet/IndigoObject.cs
@@ -1694,6 +1694,11 @@ namespace com.epam.indigo
             return dispatcher.checkResult(IndigoLib.indigoCheckObj2(self, type));
         }
 
+        public IndigoObject fingerprint()
+        {
+            return fingerprint("");
+        }
+
         public IndigoObject fingerprint(string type)
         {
             if (type == null)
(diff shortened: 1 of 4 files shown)


MysterionRise (approved):
Looks fine by me. Need to introduce CI for .NET tests later

mkviatkovskii (approved):

api/dotnet/IndigoObject.cs
```diff
@@ -1694,6 +1694,11 @@ public string check2(string type)
             return dispatcher.checkResult(IndigoLib.indigoCheckObj2(self, type));
         }
 
+        public IndigoObject fingerprint()
```
SunFellow, line 1697: Do we have real demand for such method (except in test, where no problem to pass null)?

api/tests/dotnet/IndigoTest.cs
```diff
@@ -17,6 +17,18 @@ public void TestIndigoVersion()
...
+        [TestMethod]
+        public void TestIndigoGetOneBitsList()
+        {
+            var indigo = new Indigo();
+            var indigoObject = indigo.loadMolecule("C1=CC=CC=C1");
+
+            Assert.AreEqual(
+                "1698 1719 1749 1806 1909 1914 1971 2056",
+                indigoObject.fingerprint().oneBitsList(),
+                "same one bits as in string 1698 1719 1749 1806 1909 1914 1971 205");
```
SunFellow, line 29: Why missed the last `6` in the string? (`205` instead `2056`)
Here and in the referred Java pull request (apparently copy-pasted).

f1nzer: Yes, my bad.
Imho this data shouldn't be mentioned in the message parameter.

PR #334, opened 2021-02-19
Refactoring/molecule ionize

zproksi:
better to watch both commits separately.


mkviatkovskii (approved):

molecule/src/molecule_ionize.cpp
```diff
@@ -444,433 +419,423 @@ void MoleculePkaModel::_loadSimplePkaModel()
     _model.simple_model_ready = true;
 }
 
-void MoleculePkaModel::_loadAdvancedPkaModel()
+namespace  // only for data
```
SunFellow, line 422: The comment is vague (especially it includes a function). Better specify the purpose or remove any comment.

SunFellow, line 422: This namespace (by its designation) should be inner of `namespace indigo`, but that one is only referred by `using namespace indigo`.

molecule/src/molecule_ionize.cpp
```diff
@@ -444,433 +419,423 @@ void MoleculePkaModel::_loadSimplePkaModel()
...
+        {"7300021320000|166000206220006400020421000|640002042100073000213200008200022201000820002220100073000213110007300021311000", 7.40f, 0.00f},
+        {"82-10023110000|6400020421000|640002043000082-100231100008200022201000", 3.40f, 0.00f},
+        {"7300021320000|166000206220006400020421000|640002042100073000213200008200022201000820002220100064000204210008200022201000", 4.57f, 0.00f},
+        {"7300021320000|64000204210007300021330000|73000213200007300021320000820002220100064000204200006400020420000", 3.60f, 0.00f},
+        {"7300021320000|64000204210006400020421000|73000213110007300021320000820002220100064000204210008200022201000", 9.69f, 0.00f},
+        {"82-10023110000|6400020421000|640002042100082-100231100008200022201000", 3.20f, 0.00f}
+    };
 
-    for (auto i = 0; i < NELEM(advanced_pka_model); i++)
+    void LoadPkaDefToModel(RedBlackStringObjMap<Array<float>>& adv_model, PkaDef* from, PkaDef* const end)
```
SunFellow, line 820: It's better to have iterators of the same type - both `const` in this case; not to "save on" intermediate variables. :)

molecule/src/molecule_ionize.cpp
```diff
@@ -444,433 +419,423 @@ void MoleculePkaModel::_loadSimplePkaModel()
...
-        {PKA_BASIC, "7300021330000|640002041000064000204200006400020430000|7300021330000640002042000064000204100006400020420000", 10.21f, 0.01f},
-        {PKA_BASIC, "7300021330000|640002042000064000204200006400020420000|6400020410000730002133000064000204200006400020420000", 9.06f, 1.39f},
-        {PKA_BASIC, "7300021330000|640002042000064000204200006400020420000|6400020420000730002133000064000204200006400020421000", 8.35f, 0.00f},
-        {PKA_BASIC, "7300021320000|64000204300006400020430000|64000204100006400020410000730002132000064000204100006400020410000", 11.05f, 0.00f},
-        {PKA_BASIC, "7300021330000|640002042000064000204200006400020420000|6400020410000730002133000064000204100006400020410000", 10.75f, 0.00f},
-        {PKA_BASIC, "7300021330000|640002042000064000204200006400020420000|6400020420000730002133000064000204200006400020420000", 8.50f, 1.97f},
-        {PKA_BASIC, "7300021320000|1440002044000014400020440000|6400020410000640002041000064000204100007300021320000640002041000064000204100006400020410000",
+    // These are basics
+    PkaDef advanced_pka_model_basic[] = {
```
SunFellow, line 432: Let it be `static const` or even `static constexpr`. (we've passed the point of no return to C++11)

molecule/src/molecule_ionize.cpp
```diff
@@ -444,433 +419,423 @@ void MoleculePkaModel::_loadSimplePkaModel()
...
+        {"7300021330000|640002042000064000204300006400020430000|640002043000073000213300006400020420000640002043000064000204200006400020430000",
          6.66f, 0.00f},
-        {PKA_BASIC, "7300021330000|640002042000064000204200006400020430000|64000204110007300021330000640002042000064000204400006400020440000", 5.40f, 0.00f},
+        {"7300021330000|640002042000064000204200006400020430000|64000204110007300021330000640002042000064000204400006400020440000", 5.40f, 0.00f},
     };
 
-    _model.adv_a_pkas.clear();
-    _model.adv_b_pkas.clear();
+    // these are Acids
+    PkaDef advanced_pka_model_acid[] = {
```
SunFellow, line 696: Let it be `static const` or even `static constexpr`.

molecule/src/molecule_ionize.cpp
SunFellow, line 820: Moreover, since the very data do not change, the params can be:
```
const PkaDef* const from, const PkaDef* const end
```

molecule/src/molecule_ionize.cpp
```diff
@@ -444,433 +419,423 @@ void MoleculePkaModel::_loadSimplePkaModel()
...
+namespace  // only for data
 {
-    enum PkaType
-    {
-        PKA_ACID,
-        PKA_BASIC
-    };
-
-    struct _PkaDef
+    struct PkaDef
```
SunFellow, line 424: It also makes sense to reuse this type and refactor this:
```
    static _PkaDef simple_pka_model[] = {...
```
inside `MoleculePkaModel::_loadSimplePkaModel()`

SunFellow (approved):
Just small proposals.

molecule/src/molecule_ionize.cpp
zproksi: This data/namespace is for the local usage only.

molecule/src/molecule_ionize.cpp
zproksi: It is static data. And therefore it const anyway.

PR #335, opened 2021-02-23
Meet C++11

SunFellow:
This pr is intended to switch from oldie C/C++ constructions to C++11 standard that offers more strict type checking and many utility classes and functions.

### DIFF
--- a/api/c/bingo-nosql/src/bingo_base_index.h
+++ b/api/c/bingo-nosql/src/bingo_base_index.h
@@ -82,17 +82,17 @@ namespace bingo
         };
 
     public:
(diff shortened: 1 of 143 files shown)


common/base_c/defs.h
```diff
@@ -51,7 +74,11 @@
 #define M_PI 3.14159265358979323846
 #endif
 
-#define NELEM(arr) ((int)(sizeof(arr) / sizeof(arr[0])))
+#ifdef __cplusplus
```
mkviatkovskii, line 77: I would avoid adding C++ code to `base_c/defs.h`, since it's explicitly named as a C header.

AlexanderSavelyev:
changes were reviewed. Regression tests were run for indigo and bingo postgresql. PR can be merged

PR #340, opened 2021-02-26
class Exception:

zproksi:
~~losts virtual table,~~
 losts code field {not used}
 starts using default generated methods
Code with exception usage has been simplified.

As further options - it's seems to get rid of any child of Exception class.


common/base_cpp/exception.h
```diff
@@ -80,38 +65,16 @@ public
...
-    Namespace::ExceptionName::ExceptionName(const ExceptionName& other) : Parent()                                                                             \
-    {                                                                                                                                                          \
-        other._cloneTo(this);                                                                                                                                  \
-    }
+#define IMPL_EXCEPTION2(Namespace, ExceptionName, Parent, prefix)                            \
+    Namespace::ExceptionName::ExceptionName(const char* format, ...) : Parent(prefix ": ")   \
+    {                                                                                        \
+        va_list args;                                                                        \
+        va_start(args, format);                                                              \
+        const size_t len = static_cast<size_t>(strlen(_message));                            \
```
SunFellow, line 73: Unnecessary casting: `static_cast<size_t>` as strlen now returns `size_t`.

common/base_cpp/exception.cpp
```diff
@@ -27,81 +26,21 @@ using namespace indigo;
...
 void Exception::appendMessage(const char* format, ...)
 {
     va_list args;
-
     va_start(args, format);
 
-    char added_message[1024];
-    vsnprintf(added_message, sizeof(added_message), format, args);
-    strncat(_message, added_message, sizeof(_message));
+    const size_t len = static_cast<size_t>(strlen(_message));
```
SunFellow, line 41: Unnecessary casting: static_cast<size_t> as strlen now returns size_t.

api/plugins/bingo/src/bingo_matcher.cpp
```diff
@@ -306,9 +306,9 @@ bool BaseMatcher::_loadCurrentObject()
     }
     catch (Exception& ex)
     {
-        int db_id = _index.getIdMapping()[_current_id];
+        const int db_id = _index.getIdMapping()[_current_id];
         ex.appendMessage(" on id=%d", db_id);
-        ex.throwSelf();
+        throw;
         return false; // This statement is dummy because throwSelf always throws an exception
```
SunFellow, line 312: Let's remove this too :)

SunFellow (approved):

PR #369, opened 2021-03-22
Bugfix/365

even1024:
Calculate output formatting issue fix


mkviatkovskii:
@even1024 could you please add an assertion to the js test to make it visible what issue is fixed?

SunFellow (commented):
Couple of comments:
- It's good rule to avoid abbreviations / cut words in names of classes, methods, functions. For example, `fmtDouble2VecStr` should be named as `formatDouble2VectorString`. (Even widely used Anglicism `2` instead `To` is discussible).
If the name looks too long, than it's worth to revise the name. 
Another example, `mamMasses`. Though this is a variable, `mam` can confuse a newcomer, so it's better to repeat the long part of its counterpart function, so: `mostAbundantMasses`.

- An idea to promote OOP: when I see functions with two parameters, input and output, like `void dumpStrVector2Stream(std::vector<std::string>& vec, std::stringstream& ss)`, I want to replace them with a conversion operator. Other hand, it's bad to overload common operators, so we need a specific type. Here it can be implemented by wrapping the vector of strings to a class, like this (pseudo code):
```
class Dumper {
   const std::vector<std::string> &vec;
public:
  explicit Dumper(const std::vector<std::string> v) : vec(v);
  std::stringstream& operator << (std::stringstream& ss, Dumper d) {
        for (const auto& str : vec) {
           ... // skip details
            ss << "; ";
            ss << str;
       }
       return ss;
    }
};
```
Then unobvious call `dumpStrVector2Stream(molWeights, molecularWeightStream);` will look straightforward:
```
molecularWeightStream << Dumper(molWeights);
```

PR #371, opened 2021-03-24
#338 Fix errors in CML loader with pseudo atoms

### LINKED ISSUE #338: Convert method for API calls return wrong structure after converting CML format
**Steps to reproduce:**
call `v2/indigo/convert` for [file.zip](https://github.com/epam/Indigo/files/6043825/file.zip) to convert CML into mol

**Expected result:**
the endpoint returns the correct mol with AH atom and C atoms

**Actual result:**
the AH atom disappears from mol and replaced with a C atom in spite of the presence of AH in the CML

SunFellow:
* Loading CML format: throw an exception on pseudo atom, if not query. This forces the caller to retry loading as QueryMolecule.
* Set Pseudo Atom property also for Carbon, which `elementType` was kept "C" at previous steps for pseudo atoms: AH, QH, XH, MH, X, M.
* Python test that repeats the essential logic of indigo-service /v2/convert.
* C++ test for debugging.


mkviatkovskii:
Dear @SunFellow,
Could you please tell me the current state of this pull request? Are there any blockers?

MysterionRise:
@SunFellow is it still actual?

PR #376, opened 2021-03-25
Bugfix/370

even1024:
Fix the result format for calculate to conform remote Indigo service.

### DIFF
--- a/api/wasm/indigo-ketcher/indigo-ketcher.cpp
+++ b/api/wasm/indigo-ketcher/indigo-ketcher.cpp
@@ -402,6 +402,8 @@ namespace indigo
 
         if (str.size())
             vec.push_back(str);
+        else
+            vec.push_back("Unknown error");
     }
 
     void calculate_molecule(IndigoKetcherObject iko, std::stringstream& molecularWeightStream, std::stringstream& mostAbundantMassStream,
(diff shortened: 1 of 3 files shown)


api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -320,6 +321,61 @@ namespace indigo
         iko.set(mid, IndigoKetcherObject::EKETMolecule);
     }
 
+    class VectorString
```
SunFellow, line 324: Since there is `VectorStringBracketed`, it makes sense to name this one as `VectorStringSemicolonSeparated` or laconic `VectorStringSemicoloned` :)

api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -333,7 +389,7 @@ namespace indigo
         }
     }
 
-    void fmtDouble2VecStr(double arg, std::vector<std::string>& vec)
+    void pushDoubleAsStr(double arg, std::vector<std::string>& vec)
```
SunFellow, line 392: (This is about line 403 - cannot attach comment to old unchanged line):
```
if (str.size())
```
I doubt that we should anyway skip adding the element to the vector. 
Once the function is called, the corresponding "cell" needs to be added even with empty string.

SunFellow (changes requested):
Generic: This pull request is on top of previous #375.
However two commits are completely independent, so they can be merged from independent branches. 
Proposal: split pull requests, i.e. remove the first commit from this one.

SunFellow (approved):
Well, drop splitting #375 and #376.
However, in this case follow the order of merging: first merge #375, next #376 (this one).

Issue #352, opened 2021-03-11
MS SQL server msdb.sysxmitqueue table bloat


AlexanderSavelyev:
The msdb.sysxmitqueue table ends up with millions of records.  Our dba has put together some questions to try to track down this issue.  
1.	What is the Service Broker used for in the application?  What is the flow and purpose of the messages
2.	Is it required in our implementation?
3.	If not, Can it be disabled or removed?
4.	The msdb.sysxmitqueue system table in msdb database had over 30m queue entries and bloated the DB to 35GB.  Has this issue been reported before?
5.	If so, what was the root cause and how is it normally fixed?
6.	If not, what could potentially cause this issue - we have no message queue entries currently to help in debugging as the queue was so large it was unresponsive and needed to be reset.
7.	Does the issue happen only on some versions of SQL Server - 2017 Enterprise, 2017 Express etc?

AlexanderSavelyev:
the only place there the service broker is used
https://github.com/epam/Indigo/blob/1196b66f9f889359eab50fd705b6b8ab16c24455/bingo/sqlserver/sql/bingo_create.sql#L115
where it subscribes an event from databases, e.g. remove index if a table was removed
So a possible step to reproduce is to 
loop
  create table
  create index
  drop table

AlexanderSavelyev:
possible fix to to use clean for the queue
`declare @c uniqueidentifier
while(1=1)
begin
    select top 1 @c = conversation_handle from dbo.queuename
    if (@@ROWCOUNT = 0)
    break
    end conversation @c with cleanup
end`

thomaskoppcsharp:
-- What condition/action causes a sys.conversation_endpoints object to go from status_desc STARTED_OUTBOUND to CONVERSING? 
-- Can the event notification be created on the DATABASE level instead of the SERVER level, thus: 
	create event notification $(bingo)_$(database)_logout_notify on database

AlexanderSavelyev:
We managed to reproduce the issue. Unfortunately, there are no simple solutions, e.g. it is not possible to change the notification from server to database level since the necessary events can only be taken from the database level.
Overall the problem is that backup overwrites the database notifications
Right now we are working on updating the installation script to be able to fix the following scenarios:
* install backup without bingo on a database with bingo
* install backup with bingo on a database without bingo
* install backup with bingo on a database with bingo (with the same id)

The solution will require to execute the installation script after each backup restore. @thomaskoppcsharp  will it work for you?

[...]

PR #378, opened 2021-03-25
bingo-mssql: added checks for edge cases during installation

f1nzer:
This pull request addresses several issues mentioned in https://github.com/epam/Indigo/issues/352

1. Added DDL trigger for database drop check - otherwise, a database could be dropped manually without bingo deinstallation and the same problem will appear (with service broker events flood)
2.  It is necessary to generate new service_broker_guid when any other database has identical active service_broker_guid (target database might be used as a template and be created via backup, so that's why duplications are possible and in addition, the same service_broker_guid might be enabled on another db)
3. When the user installs backup without bingo on a database with bingo (with a new service_broker_guid) there will be a registered event notification on the previous service_broker_guid. In that case installation script now removes this event notification and installs a new one.

also added some gitignore rules for bingo-related files

### DIFF
--- a/bingo/.gitignore
+++ b/bingo/.gitignore
@@ -1,0 +1,14 @@
+sqlserver/bin
+sqlserver/obj
+
+sqlserver/sqlserver_script_generator/bin
+sqlserver/sqlserver_script_generator/obj
+
+sqlserver/sql/assembly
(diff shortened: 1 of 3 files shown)


AlexanderSavelyev:
reviewed and merged the PR

PR #379, opened 2021-03-25
bugfix/370

even1024:
Square bracket in reactions fix

### DIFF
--- a/api/wasm/indigo-ketcher/indigo-ketcher.cpp
+++ b/api/wasm/indigo-ketcher/indigo-ketcher.cpp
@@ -349,9 +349,10 @@ namespace indigo
     class VectorStringBracketed
     {
         const std::vector<std::string>& vec;
+        bool both{false};
 
     public:
-        explicit VectorStringBracketed(const std::vector<std::string>& v) : vec(v)
+        explicit VectorStringBracketed(const std::vector<std::string>& v, bool both ) : vec(v), both( both )
         {
         }
         friend std::stringstream& operator<<(std::stringstream& ss, const VectorStringBracketed& vs);
@@ -362,14 +363,12 @@ namespace indigo
         bool isFirst = true;
         for (const auto& str : vs.vec)
         {
-            if (vs.vec.size() > 1)
-            {
-                if (isFirst)
-                    isFirst = false;
-                else
-                    ss << "+";
+            if (isFirst)
+                isFirst = false;
+            else
+                ss << "+";
+            if( vs.vec.size() > 1 || vs.both )
                 ss << "[" << str << "]";
-            }
             else
                 ss << str;
         }
@@ -468,11 +467,10 @@ namespace indigo
         grossFormulaStream << VectorStringSemicoloned(grossFormulas);
     }
 
-    void calculate_iteration_object(const IndigoObject& iterator, std::stringstream& molecularWeightStream, std::stringstream& mostAbundantMassStream,
-                                    std::stringstream& monoisotopicMassStream, std::stringstream& massCompositionStream, std::stringstream& grossFormulaStream,
+    void calculate_iteration_object(const IndigoObject& iterator, std::vector<std::string>& molWeights, std::vector<std::string>& mamMasses,
+                                    std::vector<std::string>& misoMasses, std::vector<std::string>& massCompositions, std::vector<std::string>& grossFormulas,
                                     const std::vector<int>& selected_atoms, int& base)
     {
-        std::vector<std::string> molWeights, mamMasses, misoMasses, massCompositions, grossFormulas;
         while (const auto id = _checkResult(indigoNext(iterator.id)))
         {
             auto mol = IndigoKetcherObject(id, _checkResult(indigoCheckQuery(id)) ? IndigoKetcherObject::EKETMoleculeQuery : IndigoKetcherObject::EKETMolecule);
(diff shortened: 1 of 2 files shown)


api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -362,14 +363,12 @@ namespace indigo
...
-            {
-                if (isFirst)
-                    isFirst = false;
-                else
-                    ss << "+";
+            if (isFirst)
+                isFirst = false;
+            else
+                ss << "+";
+            if( vs.vec.size() > 1 || vs.both )
```
SunFellow, line 370: This expression is invariant in the cycle and can be calculated outside `for`.
Moreover, the only case when it's false (when size == 1) does not require cycle `for` at all.
So,
```
if (vs.vec.size() == 1 && !vs.both)
{
   ss << str;
   return;
}
```
// otherwise for

api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -503,44 +501,50 @@ namespace indigo
...
-        monoisotopicMassStream << VectorStringBracketed(misoMasses);
-        massCompositionStream << VectorStringBracketed(massCompositions);
-        grossFormulaStream << VectorStringBracketed(grossFormulas);
     }
 
     void calculate_reaction(const IndigoKetcherObject& iko, std::stringstream& molecularWeightStream, std::stringstream& mostAbundantMassStream,
                             std::stringstream& monoisotopicMassStream, std::stringstream& massCompositionStream, std::stringstream& grossFormulaStream,
                             const std::vector<int>& selected_atoms)
     {
+        std::vector<std::string> molWeights[2], mamMasses[2], misoMasses[2], massCompositions[2], grossFormulas[2];
```
SunFellow, line 510: To make design clear, let's introduce, for named idexes for products and reactants, example:
```
enum { IDX_PRODUCTS = 0, IDX_REACTANTS = 1 };
```
and use them instead [0] and [1]. 
Then it will give a clue why each vector is 2-ice and what are their roles.

api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -503,44 +501,50 @@ namespace indigo
...
-        std::stringstream mws, mams, misos, mcs, gfs;
-        calculate_iteration_object(IndigoObject(_checkResult(indigoIterateProducts(iko.id()))), mws, mams, misos, mcs, gfs, selected_atoms, base);
-        if (grossFormulaStream.str().size() && gfs.str().size()) // check if we have gross formula for both reagents and products
-        {
+        calculate_iteration_object(IndigoObject(_checkResult(indigoIterateReactants(iko.id()))), molWeights[0], mamMasses[0], misoMasses[0], massCompositions[0], grossFormulas[0], selected_atoms, base);
+        calculate_iteration_object(IndigoObject(_checkResult(indigoIterateProducts(iko.id()))), molWeights[1], mamMasses[1], misoMasses[1], massCompositions[1], grossFormulas[1], selected_atoms, base);
+        
+        if (grossFormulas[0].size() && grossFormulas[1].size()) // check if we have gross formula for both reagents and products
+        {
+            molecularWeightStream << VectorStringBracketed(molWeights[0], true ); // true means we have both: products and reagents so all components are bracketed
```
SunFellow, line 524: Please move the comment to separate line at the beginning of the block.
Also please add similar comment at the beginning of the else block.

api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -503,44 +501,50 @@ namespace indigo
...
+            grossFormulaStream << VectorStringBracketed(grossFormulas[1], true );
+        } else if( grossFormulas[0].size() || grossFormulas[1].size() )
+        {
+            int idx = grossFormulas[0].size() ? 0 : 1;
+            molecularWeightStream << VectorStringBracketed(molWeights[idx], false );
+            mostAbundantMassStream << VectorStringBracketed(mamMasses[idx], false );
+            monoisotopicMassStream << VectorStringBracketed(misoMasses[idx], false );
+            massCompositionStream << VectorStringBracketed(massCompositions[idx], false );
+            grossFormulaStream << VectorStringBracketed(grossFormulas[idx], false );
         }
```
SunFellow, line 547: Don't we need to output something when both products and reactants values are empty?

api/wasm/indigo-ketcher/indigo-ketcher.cpp
```diff
@@ -349,9 +349,10 @@ namespace indigo
     class VectorStringBracketed
     {
         const std::vector<std::string>& vec;
+        bool both{false};
```
SunFellow, line 352: Consider to move from single vector + both to pair of vectors.
Since we need bracketed object only for reactions, it supposes two parts. Let's model "two parts" not outside this class by vector[2], but inside this specific class. In this case, the caller can use: 

instead `molWeights[0]` -> `molWeights.products`
instead `molWeights[1]` -> `molWeights.reactants`

mkviatkovskii (changes requested):
I don't see any new tests here. Could you please add some, like test for `check()`? Testing via Ketcher is obviously not a best way to do.

api/wasm/indigo-ketcher/indigo-ketcher.cpp
even1024: fixed

api/wasm/indigo-ketcher/indigo-ketcher.cpp
even1024: fixed

api/wasm/indigo-ketcher/indigo-ketcher.cpp
even1024: done

api/wasm/indigo-ketcher/indigo-ketcher.cpp
even1024: no, we don't need.

api/wasm/indigo-ketcher/indigo-ketcher.cpp
even1024: done

mkviatkovskii (approved):

PR #387, opened 2021-04-03
cdx & check tests' fail fix

even1024:
cdx files restored


core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -926,7 +926,7 @@ int BaseMolecule::getAtomRadical_NoThrow(int idx, int fallback)
     {
         return getAtomRadical(idx);
     }
-    catch (Element::Error&)
+    catch (Exception& ex)
```
SunFellow, line 929: Better use `const Exception &`.

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -938,7 +938,7 @@ int BaseMolecule::getAtomValence_NoThrow(int idx, int fallback)
     {
         return getAtomValence(idx);
     }
-    catch (Element::Error&)
+    catch (Exception& ex)
```
SunFellow, line 941: Better use `const Exception &`

PR #394, opened 2021-04-08
Add more specific reaction test files

khyurri:
This PR fixes bug with indigo reaction java failing test

Test checks that CMF returned by EuclidSimilarityMatch (same TverskySimilarityMatch) will be similar with target. In some cases it is not. 

This fix just selects more specific reaction file as target, not random one.

### DIFF
--- a/api/plugins/bingo-elastic/java/src/test/java/com/epam/indigo/predicate/ReactionsTest.java
+++ b/api/plugins/bingo-elastic/java/src/test/java/com/epam/indigo/predicate/ReactionsTest.java
@@ -3,16 +3,19 @@ package com.epam.indigo.predicate;
 import com.epam.indigo.BaseElasticTest;
 import com.epam.indigo.model.Helpers;
 import com.epam.indigo.model.IndigoRecordReaction;
+import com.epam.indigo.model.fields.FieldNotFoundException;
 import org.junit.jupiter.api.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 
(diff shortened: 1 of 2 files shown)


api/plugins/bingo-elastic/java/src/test/java/com/epam/indigo/predicate/ReactionsTest.java
```diff
@@ -75,18 +83,53 @@ public void similarityMatchTest() {
...
+        if (targetReaction == null) {
+            Assertions.fail(new Exception("Target reaction not found"));
+        }
+        return targetReaction;
+    }
+
     @Test
     public void euclidMatchTest() {
-        IndigoRecordReaction targetReaction = reactions.get(3);
+        IndigoRecordReaction targetReaction = getTargetReaction("42837.rxn");
```
SunFellow, line 107: How about to switch all other tests to `getTargetReaction(<corr. file>)` ?

api/plugins/bingo-elastic/java/src/test/java/com/epam/indigo/predicate/ReactionsTest.java
```diff
@@ -41,15 +44,20 @@ protected void check(IndigoRecordReaction target, Iterable<IndigoRecordReaction>
...
+                    reaction.addCustomObject("File", path.toString());
                     reactions.add(reaction);
                 }
         );
+        System.out.println("Loaded reactions count:");
+        System.out.println(reactions.size());
         Iterable<IndigoRecordReaction> reactionIterable = reactions;
         repositoryReaction.indexRecords(reactionIterable, reactions.size());
-        TimeUnit.SECONDS.sleep(5);
+        TimeUnit.SECONDS.sleep(15);
```
SunFellow, line 60: Is any Sleep necessary at all? Where is asynchrony here?

MysterionRise: @SunFellow that's eventual consistency of Elastic plays in. One of the idea for improvements was to allow have method on repository so it would constantly check that actually documents got indexed. @khyurri @mkviatkovskii do we think it's reasonable thing to add to the API?

PR #404, opened 2021-04-26
Draft: #405: api: added unit tests via gtest

mkviatkovskii:


AlexanderSavelyev:
@mkviatkovskii  I would suggest to merge the tests which are passed and create additional tickets to resolve the memory issues. Having unit tests in the repo will be very useful.

PR #405, opened 2021-04-28
Reactions support for ket-format #396

even1024:
Feature implementation:
https://github.com/epam/Indigo/issues/396


core/indigo-core/molecule/src/molecule_json_loader.cpp
```diff
@@ -333,19 +328,19 @@ void MoleculeJsonLoader::parseBonds(const rapidjson::Value& bonds, BaseMolecule&
...
+                                            const Value& subs = b["subs"];
+                                            if (subs.Size() == 4)
+                                            {
+                                                for (int k = 0; k < 4; ++k)
+                                                {
+                                                    substituents[k] = subs[k].GetInt();
+                                                }
+                                            }
+                                            mol.cis_trans.add(bond_idx, substituents, stereo == 7 ? MoleculeCisTrans::CIS : MoleculeCisTrans::TRANS);
+                                        }*/
```
mkviatkovskii, line 343: Do we need this commented code?

even1024: removed

utils/indigo-depict/main.c
```diff
@@ -922,11 +922,13 @@ int main(int argc, char* argv[])
         _prepare(obj, p.aromatization);
         if (p.action == ACTION_LAYOUT)
         {
-            indigoLayout(obj);
+            // indigoLayout(obj);
```
mkviatkovskii, line 925: Why is this line commented? It seems like we need to layout the molecule

even1024: It breaks existing layout, so the line is optional.

even1024: fixed

core/indigo-core/reaction/src/reaction_json_saver.cpp
```diff
@@ -16,15 +16,43 @@
...
 #include "reaction/reaction.h"
+#include "reaction/reaction_json_saver.h"
 
 using namespace indigo;
+using namespace indigo;
+using namespace rapidjson;
 
 IMPL_ERROR(ReactionJsonSaver, "reaction KET saver");
 
+void ReactionJsonSaver::_getBounds(BaseMolecule& mol, Vec2f& min, Vec2f& max, float scale)
```
SunFellow, line 36: Better to avoid adding names `min` and `max` (literally), because Windows headers historically defined macros `min` and `max`, and we don't use consistent preinclude header.
The same refers to methods `Vec2f::min()` and `Vec2f::max()`, but it's subject of refactoring.

utils/indigo-depict/main.c
SunFellow: are you sure? (here's another clause and the same action)

even1024: Both indigoLayout(obj) was commented for testing purposes to prevent breaking of existing layout in the ket-files.

PR #406, opened 2021-04-29
IndigoCheck2 and IndigoCheckObject2 refactored

even1024:
Old checks removed from indigo


api/c/indigo/src/indigo_misc.cpp
```diff
@@ -1419,118 +1417,32 @@ CEXPORT const char* indigoCheckObj2(int item, const char* check_flags)
...
+        }
+        catch (Exception& e)
+        {
+            out.printf("{\"LOAD\":{\"message\":\"Error at loading structure. %s\"}}", e.message());
+        }
+
+        out.writeChar(0);
+        return tmp.string.ptr();
+    }
+    INDIGO_END(0);
```
mkviatkovskii, line 1443: Better use `nullptr` here.

api/c/indigo/src/indigo_structure_checker.h
```diff
@@ -43,8 +43,8 @@ namespace indigo
         CheckResult check(const IndigoObject& item, const std::vector<CheckTypeCode>& check_types = std::vector<CheckTypeCode>(),
                           const std::vector<int>& selected_atoms = std::vector<int>(), const std::vector<int>& selected_bonds = std::vector<int>());
 
-        std::string toJson(const StructureChecker2::CheckResult& res);
-        std::string toJson2(const StructureChecker2::CheckResult& res);
+        std::string toJson(const StructureChecker::CheckResult& res);
+        std::string toJson2(const StructureChecker::CheckResult& res);
```
mkviatkovskii, line 47: Do we still need both `toJson` and `toJson2` methods?

api/c/indigo/src/indigo_misc.cpp
even1024: This is not nullptr. Here we return pointer to the null-terminated string. 0 here - means 0x0 character.

api/c/indigo/src/indigo_structure_checker.h
even1024: fixed

api/c/indigo/src/indigo_misc.cpp
SunFellow: @even1024, you are mistaken. 0 is the value returned by the function. Function return type is `const char *`. This means it returns null-pointer. 
If you want to return empty string, type: `INDIGO_END("");`

Issue #433, opened 2021-06-17
api: integration tests: substructure/tau_sub unmatches two cases


mkviatkovskii:
Reproduces on branch `indigo-1.5`. Steps to reproduce:
1. Run `indigo-integration-tests` with `-p substructure/tau_sub.py` key.
2. See the `api/tests/integration/out/substructure/tau_sub.py.diff` file. It should be like this:
```
- matched
+ unmatched(unexpected)
- matched
+ unmatched(unexpected)
```
It means that two substructure matching procedures are expected to match, but are not matched. You need to:
1. Understand what match procedures failed (probably by adding debug prints to the Python code).
2. Take a look at compounds and understand if they should match or not (ask @mkviatkovskii, @AlexanderSavelyev or @IuriiPuzanov if needed).
3. Either fix the substructure or (if change if correct) update the reference file (copy `api/tests/integration/out/substructure/tau_sub.py.out to api/tests/integration/ref/substructure/tau_sub.py.ouy`.

Python code of the test could be found at `api/tests/integration/tests/substructure/tau_sub.py`.

khyurri:
`substructure/tau_sub.py` contains 2 failed tests. 

Tested on 1.4 (current release): 

1. 
```
>>> mol1 = indigo.loadQueryMolecule("CC1(C)NC(=O)C2=CC=CC=C2N1")
>>> mol2 = indigo.loadMolecule("CC(=C)NC1=CC=CC=C1C(N)=O")
>>> testSingleTauMatch(mol1, mol2, "TAU R-C", True, False)
unmatched(unexpected)
```
**Test also failed. NOT A REGRESSION**

2.
```
>>> mol1 = indigo.loadQueryMolecule("CC1(C)NC(=O)C2=CC=CC=C2N1")
>>> mol2 = indigo.loadMolecule("CC(=C)NC1=CC=CC=C1C(N)=O")
>>> mol2.aromatize()
>>> testSingleTauMatch(mol1, mol2, "TAU R-C", True, False)
testSingleTauMatch(mol1, mol2, "TAU R-C R2", True, False)
```

Molecule:
![Screen Shot 2021-06-18 at 9 04 43 PM](https://user-images.githubusercontent.com/4924485/122600682-d39f0880-d078-11eb-9a30-510c1c0a638a.png)

Query: 
[...]

PR #439, opened 2021-06-26
indigo: move tau ring-chain test into todo

AlexanderSavelyev:
Move TAU R-C test into todo list for #433.


SunFellow (approved):
The very changes are OK and correspond the subject.

AlexanderSavelyev:
close since the tests were fixed

PR #440, opened 2021-06-26
Indigo 1.5 bugfixes/bug431

even1024:


api/c/indigo/src/indigo_internal.h
```diff
@@ -200,8 +200,7 @@ class DLLEXPORT IndigoObject
 
 protected:
     AutoPtr<Array<char>> _dbg_info; // allocated by debugInfo() on demand
-private:
-    IndigoObject(const IndigoObject&);
+    IndigoObject(const IndigoObject&){};
```
SunFellow, line 203: Doubtful change. If we want to have copy-constructor, it's better to leave it for compiler by default. If we want to override the default constructor, at least `_type` should be copied.

core/indigo-core/common/base_cpp/array.h
```diff
@@ -25,6 +25,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <utility>
+#include <list>
```
SunFellow, line 28: IMHO, this change can confuse further refactorings in perspective of removing `array.h`. Better insert `#include <list>` where the list is used.

api/c/indigo/src/indigo_internal.h
even1024: Missed the 'type' field. Fixed.

Issue #431, opened 2021-06-17
api: integration tests: deco/deco_sdf segfaults or reports "array: invalid index 0 (size=0)"


mkviatkovskii:
Could be reproduced on branch `indigo-1.5`. To reproduce:
1.  Uncomment two lines in `api/tests/integration/tests/deco/deco_sdf.py` after `TODO` comment.
2. Debug `indigo-integration-tests` with flag `-p deco/deco_sdf`.

On Linux segfault happens, on Windows and macOS only `array: invalid index 0 (size=0)` uncaught exception appears.

PR #442, opened 2021-06-30
#431 issue fix

even1024:
Removing references from BaseMolecule class because they don't get copied with shallow copy.


api/c/indigo/src/indigo_layout.cpp
```diff
@@ -70,8 +70,8 @@ CEXPORT int indigoLayout(int object)
                 mol->clearBondDirections();
                 try
                 {
-                    mol->stereocenters.markBonds();
-                    mol->allene_stereo.markBonds();
+                    mol->stereocentersMarkBonds();
+                    mol->allene_stereoMarkBonds();
```
mkviatkovskii, line 74: I suggest using the same naming style, `alleneStereoMarkBonds`

api/c/indigo/src/indigo_misc.cpp
```diff
@@ -1032,7 +1032,7 @@ CEXPORT int indigoNormalize(int structure, const char* options)
         {
             // Validate cs-trans because it can disappear
             // For example: [O-]/[N+](=C\C1C=CC=CC=1)/C1C=CC=CC=1
-            mol.cis_trans.validate();
+            mol.cis_transValidate();
```
mkviatkovskii, line 1035: Same here, like `validateCisTrans`

api/c/tests/unit/basic/deco_sdf.cpp
```diff
@@ -0,0 +1,140 @@
...
+        testScaffold(dataPath("molecules/basic/thiazolidines.sdf"), "approx", false);
+        testScaffold(dataPath("molecules/basic/thiazolidines.sdf"), "approx 3", false);
+        testScaffold(dataPath("molecules/basic/sugars.sdf"), "exact", true);
+        testScaffold(dataPath("molecules/basic/sugars.sdf"), "approx", false);
+    }
+    catch (Exception& e)
+    {
+        ASSERT_STREQ("", e.message());
+    }
+}
```
mkviatkovskii, line 140: I don't think we need this test here - C++ unittests should be small and fast. We have this test in integration tests.

api/c/indigo/src/indigo_misc.cpp
even1024: fixed

api/c/tests/unit/basic/deco_sdf.cpp
even1024: test removed

api/c/indigo/src/indigo_layout.cpp
even1024: fixed

PR #481, opened 2021-09-08
core: use standard C++11 threading structures instead of own implementations

vdksoft:
Refactored all synchronization primitives except thread-local storage.


core/render2d/render_context.h
```diff
@@ -166,35 +166,29 @@ namespace indigo
...
             ~TextLock()
             {
-#ifdef _WIN32
-                osMutexDelete(&_mutex);
-#endif
             }
 
             void lock()
             {
 #ifdef _WIN32
```
SunFellow, line 177: Probably, leaving `#ifdef _WIN32` can misuse the code in the future.

core/render2d/render_context.h
```diff
@@ -166,35 +166,29 @@ namespace indigo
...
-                osMutexUnlock(&_mutex);
+                _mutex.unlock();
 #endif
             }
 
         private:
 #ifdef _WIN32
-            os_mutex _mutex;
+            std::mutex _mutex;
```
SunFellow, line 191: It makes sense to add `#include <mutex>` to this header.

core/render2d/render_context.h
SunFellow: However, it's not a subject of refactoring pull-request...

core/indigo-core/common/base_cpp/os_sync_wrapper.h
```diff
@@ -19,59 +19,17 @@
...
-        {
-            if (_lock != NULL)
-                _lock->Unlock();
-        }
-
-    private:
-        T* _lock;
-    };
-    typedef OsLockerT<OsLock, false> OsLocker;
-    typedef OsLockerT<OsLock, true> OsLockerNullable;
```
SunFellow, line 71: Replace OsLockerNullable with std::unique_lock. (64d47b47bcb335575f1c46964975c21832ef0f6d)
resulted in increasing code in 6 places. So, some macro/template instead this would be useful.

core/indigo-core/common/base_cpp/os_sync_wrapper.h
```diff
@@ -113,9 +74,9 @@ namespace indigo
     // Thread-safe static local variables initialization object
     //
 
-    DLLEXPORT OsLock& osStaticObjConstructionLock();
+    DLLEXPORT std::mutex& osStaticObjConstructionLock();
```
SunFellow, line 77: It looks like this function is nowhere used. We can remove it.

SunFellow:
Consider using `std::recursive_mutex` since we can't make sure that users of SDK properly use synchronization classes and wrappers.

PR #502, opened 2021-10-08
New indigo service alpha preview with JSON:API support

khyurri:
This PR adds new indigo service with JSON:API interface

- [x] Add README or documentation
- [x] Add tests for all methods
- [x] Add pylint and mypy to pre-commit
- [x] Add CI/CD


api/http/indigo_service/jsonapi.py
```diff
@@ -0,0 +1,570 @@
...
+
+# Formats
+
+
+class CompoundFormat(str, Enum):
+    AUTO = "auto"
+    MOLFILE = "molfile"
+    SMILES = "smiles"
+    CML = "cml"
+    SMARTS = "smarts"
```
mkviatkovskii, line 43: We can also support Ket format to let Ketcher team try new service.

api/http/indigo_service/indigo_http.py
```diff
@@ -0,0 +1,216 @@
...
+            response.headers["Content-Type"] = RESP_HEADER_CONTENT_TYPE
+        return response
+
+
+@app.exception_handler(IndigoException)
+def unicorn_exception_handler(
+    request: Request, exc: IndigoException  # pylint: disable=unused-argument
+) -> JSONResponse:
+    error = jsonapi.make_error_response(exc)
+    return JSONResponse(status_code=400, content=error.dict())
```
mkviatkovskii, line 53: Should we also add a general exception handler for `Exception` and return 500 in that case?

mkviatkovskii (approved):
Looks very promising.

api/http/mypy.ini
```diff
@@ -0,0 +1,11 @@
+[mypy]
+python_version = 3.9
+strict = True
```
mkviatkovskii, line 3: Looks a little bit strange — strict mode is on, but I don't see `# typing: ignore` directives, which should be required at least for Indigo API, since it's untyped at the moment. Am I missing something here?

api/http/indigo_service/indigo_http.py
```diff
@@ -0,0 +1,216 @@
...
+    compound, *_ = service.extract_compounds(source(request))
+    target_compounds = service.extract_compounds(targets(request))
+    source_fp = compound.fingerprint("sim")
+    result = []
+    for target in target_compounds:
+        bits = indigo().commonBits(source_fp, target.fingerprint("sim"))
+        result.append(bits)
+    return jsonapi.make_common_bits_response(result)
+
+
```
mkviatkovskii, line 206: Is it complicated to add missing methods for required by Ketcher? It seems like the following are required:
* `aromatize`
* `dearomatize`
* `layout`
* `clean2d`
Also `render` is quite important, but it could be done later.

api/http/indigo_service/jsonapi.py
mkviatkovskii: And `InCHi` as well.

mkviatkovskii (approved):

api/http/mypy.ini
khyurri: I'm just turning off all indigo imports using `ignore_missing_imports = True` 
https://github.com/epam/Indigo/pull/502/files/0e7dc9f30fe65ef0e93326eb92d6791acdc2549c#diff-fe5c8c88597291fc7f9219d3411417e39b5da69faf2d8b748cf2c6b882cab8f0R7

api/http/indigo_service/jsonapi.py
khyurri: InCHi added

api/http/indigo_service/indigo_http.py
khyurri: Added `aromatize`, `dearomatize`, `clean2d` to /indigo/convert endpoint. 

`layout` and `render` needs clarification, let's move it to feature request?

api/http/indigo_service/jsonapi.py
khyurri: Added KET format. Unfortunately current CompondModel implementation supports structure only as string objects, not dict. It should be fixed in next releases.

api/http/indigo_service/indigo_http.py
khyurri: `layout` added. `render` will be added in the next release

api/http/indigo_service/indigo_http.py
khyurri: 500 by Exception is a default behaviour

Issue #539, opened 2021-11-24
Implement salt checking procedure


mkviatkovskii:
**Description**
We need to check if a chemical compound contains salt. Is should be a method that returns a boolean value:
```python
class IndigoObject:
    ...

    def checkSalt(self) -> bool:
        ...
```

**Implementation details**
We have a substructure matcher that allows us to check if a chemical compound is a part of other chemical compound. The example of how it works:
```python
from indigo import Indigo

# Create a new Indigo session
indigo = Indigo()

# Load target compound that we need to check for salts
target_molecule = indigo.loadMolecule("CC.[Na]Cl")

# Iterate over disconnected fragments of the compound, "CC" and "[Na]Cl" in this case
for target_fragment in target_molecule.iterateComponents():
    # Create a molecule from fragment, for technical reasons
    target_fragment = target_fragment.clone()
    
    # Check if current fragment is "C"
    query_molecule_c = indigo.loadMolecule("C")
    if indigo.exactMatch(target_fragment, query_molecule_c):
        print("C matched")

    # Check if current fragment is "NaCl"
    query_molecule_nacl = indigo.loadMolecule("[Na]Cl")
    if indigo.exactMatch(target_fragment, query_molecule_nacl):
        print("NaCl matched")
```

So the salt checking procedure could be implemented by calling a multiple substructure checking for different types of salts, and main challenge is to prepare the list of salts. [This list](https://github.com/rdkit/rdkit/blob/b208da471f8edc88e07c77ed7d7868649ac75100/Data/Salts.txt) could be used as a source of inspiration.

Note: to import indigo, you can install Indigo from PyPi by using following command:
```
pip install epam.indigo
```

AATDev21:
I'm a bit confused with what exactly should be recognized as "salt". Which types of fragments, individual ions or ion-counterion pairs, are supposed to be found?

For example we have a system with large organic ion and small counter ion - cetylpyridinium chloride, CCCCCCCCCCCCCCCC[N+]1=CC=CC=C1.[Cl-]. Is it correct to recognized [Cl-] as "salt" or it is necessary to have in here some [Na+] (or any other small cation) as well and mark them both as NaCl?

AlexanderSavelyev:
closing the ticket

PR #549, opened 2021-12-10
api: python: checkSalt() implemented #539

AATDev21:
- Added new IndigoObject method checkSalt with a list of basic ions for search.
- Added unittest for this method.
- Fixed typos in structure_checker.cpp, test.js and ref basic_check.py.out.


MysterionRise (changes requested):
@AATDev21  why we need basic_check.py.out file in Git?


Could you also run black + isort + pylint on the fresh additions? It looks like some of the PEP-8 guidelines are missing here

api/python/tests/test_indigo.py
```diff
@@ -9,3 +9,7 @@ def test_aromatize_smiles(self) -> None:
         m = self.indigo.loadMolecule('C1=CC=CC=C1')
         m.aromatize()
         self.assertEqual('c1ccccc1', m.smiles())
+
+    def test_check_salt(self) -> None:
+        m = self.indigo.loadMolecule('[Na+].C')
+        self.assertTrue(m.checkSalt())
```
mkviatkovskii, line 15: Let's add a couple more tests:
1. To return `True` if compound contains multiple salts.
2. To return `False` if compound does not contain any salts.
3. To return `False` if compound contains, some salt-like ion, but not as standalone fragment.

mkviatkovskii (changes requested):
Looks good, let's add a couple more tests and the merge.

mkviatkovskii:
@MysterionRise, we keep *.out file as a reference in our own-written mind-blowing test engine. Not very intuitive, but it's how it works for now.

MysterionRise:
@mkviatkovskii that's weird, but okay :)

api/python/tests/test_indigo.py
AATDev21: I'll add them, thank you!

AATDev21:
@MysterionRise , I didn't change or add any imports in here, but I'll check. Thank you!

MysterionRise:
@AATDev21 im more on the side of 2 lines between methods, etc

AATDev21:
@MysterionRise I just found a link [https://www.python.org/dev/peps/pep-0008/#blank-lines](url) with PEP8, in which one blank line is recommended for class methods separation. I could use extra lines if my method was entirely different from others in the class, but it just continued a large group of already existed molecule methods.

mkviatkovskii (approved):

Issue #533, opened 2021-11-22
Error when user tries to use Layout


ElenaOdnoshivkina:
**Steps to reproduce:**

1. Execute /v2/indigo/layout

options: {
        "smart-layout": true,
        "ignore-stereochemistry-errors": true, 
        "mass-skip-error-on-pseudoatoms": false,
        "gross-formula-add-isotopes": true,
        "gross-formula-add-rsites": true
},
output_format: "chemical/x-indigo-ket",
struct: [struct.txt](https://github.com/epam/Indigo/files/7581843/struct.txt)

**Expected Result:**
Response is structure that changed its position

**Actual Result:**
Respons is object contains error: "IndigoException: element: can not calculate valence on C, charge 0, connectivity 6"

AllaZhbanova:
The structure is broken after /v2/indigo/layout execution for the structure

before
![2021-12-14_18h31_51](https://user-images.githubusercontent.com/93334478/146029050-d91169d3-9daa-4a33-b7c0-6ba88b288874.png)

after 
![2021-12-14_18h31_21](https://user-images.githubusercontent.com/93334478/146029067-2935aa30-92c5-4bfb-8894-3b4d793f2f1e.png)

even1024:
The structure doesn't seem to be broken. This is how layout algorithm works.

AllaZhbanova:
Retested on Ketcher v.2.3 , the behavior is the same

Issue #530, opened 2021-11-19
Error when user tries to use Check method for structure with text


ElenaOdnoshivkina:
**Steps to reproduce:**

1) Execute /v2/indigo/check

"options": {
  "smart-layout": true,
  "ignore-stereochemistry-errors": true,
  "mass-skip-error-on-pseudoatoms": false,
  "gross-formula-add-rsites": true,
  "gross-formula-add-isotopes": true
    }, 
struct: [struct.txt](https://github.com/epam/Indigo/files/7568831/struct.txt),
"types": [
  "radicals",
  "pseudoatoms",
  "stereo",
  "query",
  "overlapping_atoms",
  "overlapping_bonds",
  "rgroups",
  "chiral",
  "3d"
]
      
**Expected Result:**
Response is object with errors or empty object

**Actual Result:**
Respons is HTML document
![image](https://user-images.githubusercontent.com/89905163/142591612-896701eb-71b5-454e-baa6-5fb9fd5084e7.png)

AllaZhbanova:
Fixed, tested on 
Ketcher
version2.3.3
build at 2021-12-14T07:10:36
Indigo Toolkit
version1.6.0.6-gb2d73faeb-x86_64-linux-gnu-10.2.1

Issue #523, opened 2021-11-12
generateImageAsBase64 method doesn't work correctly in standalone mode


ViktoriiaFedotkina:
Steps to reproduce:

1. call /v2/indigo/render in standalone mode with (options, struct) :
gross-formula-add-rsites: true
ignore-stereochemistry-errors: true
mass-skip-error-on-pseudoatoms: false
render-output-format: "svg" OR "png"
smart-layout: true

Expected result:
Correct PNG or SVG image encoded in base64 should be returned

Actual result:
An incorrect image is returned

[struct.zip](https://github.com/epam/Indigo/files/7528542/struct.zip)

AllaZhbanova:
Fixed, tested Standalone ketcher v,2.4

Issue #506, opened 2021-10-26
Aromatize / Dearomatize action with Fuctional Groups work incorrectly


vitalgerko:
**Steps to reproduce:**

**Aromatize:**
Execute v2/indigo/aromatize
struct
[struct.zip](https://github.com/epam/Indigo/files/7415916/struct.zip)

{
gross-formula-add-isotopes: true
gross-formula-add-rsites: true
ignore-stereochemistry-errors: true
mass-skip-error-on-pseudoatoms: false
smart-layout: true
}

**Dearomatize:**
Execute v2/indigo/dearomatize
struct
[struct.zip](https://github.com/epam/Indigo/files/7415917/struct.zip)

{
gross-formula-add-isotopes: true
gross-formula-add-rsites: true
ignore-stereochemistry-errors: true
mass-skip-error-on-pseudoatoms: false
smart-layout: true
}

**Expected result:**
The structure isn't changed. 
Indigo should not change objects that it does not know.

**Actual result:**
Indigo changes objects.

A possible suggestion for implementation: add the flag "don't change this structure" to the format and use it with a TRUE for FGs when aromatize/dearomatize is applied for ones in Ketcher.

AlexanderSavelyev:
add flag 'aromatize-skip-super-atoms'

even1024:
aromatize-skip-super-atoms added

AllaZhbanova:
The issue is still reproducible, after  aromatize applying FG is changed. Tested on
 Ketcher version2.3.3
build at 2021-12-14T07:10:36
Indigo Toolkit version1.6.0.6-gb2d73faeb-x86_64-linux-gnu-10.2.1

even1024:
Please use aromatize-skip-super-atoms=true on ketcher's side.

AllaZhbanova:
related https://github.com/epam/ketcher/issues/1108

even1024:
aromatize-skip-super-atoms renamed to -> aromatize-skip-superatoms

PR #550, opened 2021-12-13
core: multinodes support for ket format, simple objects support, #533, #530, #523, #506, #413

even1024:
multinodes support for ket format, simple objects support, #533, #530, #523, #506, #413.


core/render2d/src/render_internal.cpp
```diff
@@ -2005,6 +2008,28 @@ void MoleculeRenderInternal::_renderBonds()
...
+                break;
+                case KETSimpleObject::EKETRectangle:
+
+                break;
+                case KETSimpleObject::EKETLine:
+                break;
+            }
+        }
+    }*/
+}
```
mkviatkovskii, line 2031: This looks like a dead code.

core/indigo-core/graph/graph.h
```diff
@@ -253,6 +268,8 @@ namespace indigo
         bool _components_valid;
         int _components_count;
 
+        PtrArray<GraphMetaObject> _meta_data;
```
mkviatkovskii, line 271: Why do we use PtrArray? It should be considered as deprecated data structure now.

core/indigo-core/graph/graph.h
```diff
@@ -24,6 +24,9 @@
 #include "base_cpp/non_copyable.h"
 #include "base_cpp/obj_array.h"
 #include "base_cpp/obj_pool.h"
+#include <list>
+#include <memory>
```
mkviatkovskii, line 28: Standard library headers should be included before local headers.

mkviatkovskii: BTW, I don't see any usage of list or memory here.

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -278,6 +278,12 @@ void BaseMolecule::_mergeWithSubmolecule_Sub(BaseMolecule& mol, const Array<int>
         reaction_atom_exact_change[mapping[i]] = mol.reaction_atom_exact_change[i];
     }
 
+    if (skip_flags & COPY_BOND_DIRECTIONS)
+    {
+        _bond_directions.expandFill( edgeEnd() + mol.edgeEnd(), 0);
+    } else
+      _bond_directions.expandFill(  mol.edgeEnd(), 0);
+
```
mkviatkovskii, line 286: It's definitely required to run clang-format on changed files.

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -4122,6 +4128,34 @@ const int* BaseMolecule::getPyramidStereocenters(int idx) const
     return stereocenters.getPyramid(idx);
 }
 
+void BaseMolecule::setStereoFlagPosition(int frag_index, const Vec3f& pos)
+{
+    try
+    {
+        _stereo_flag_positions.insert(frag_index, pos);
+    }
+    catch (...)
```
mkviatkovskii, line 4137: Why do we need such generic catch? It looks like a bad code.

core/render2d/src/render_internal.cpp
even1024: fixed

core/indigo-core/graph/graph.h
even1024: We can't use std containers here

core/indigo-core/graph/graph.h
even1024: fixed

core/indigo-core/molecule/src/base_molecule.cpp
even1024: fixed

core/indigo-core/molecule/src/base_molecule.cpp
even1024: fixed

mkviatkovskii (approved):

PR #555, opened 2021-12-14
migrate to log4j 2.x

MysterionRise:


mkviatkovskii (approved):
Looks like too big time investment into deprecated code, but OK :-)

PR #570, opened 2021-12-21
ci: added static analysis and formatting checks for C++ and Python

mkviatkovskii:

PR #574, opened 2021-12-22
Some findings revealed by PR #570

SunFellow:
Reviewing changes in PR #570 reveled some points to improve or make the code laconic.

### DIFF
--- a/api/c/indigo/src/indigo_molecule.cpp
+++ b/api/c/indigo/src/indigo_molecule.cpp
@@ -36,6 +36,7 @@
 #include "molecule/query_molecule.h"
 #include "molecule/smiles_loader.h"
 #include "molecule/smiles_saver.h"
+#include <unordered_set>
 
 IndigoBaseMolecule::IndigoBaseMolecule(int type_) : IndigoObject(type_)
 {
@@ -58,8 +59,22 @@ bool IndigoBaseMolecule::is(const IndigoObject& object)
 {
     int type = object.type;
 
-    if (type == MOLECULE || type == QUERY_MOLECULE || type == REACTION_MOLECULE || type == SCAFFOLD || type == RGROUP_FRAGMENT || type == RDF_MOLECULE ||
-        type == SMILES_MOLECULE || type == CML_MOLECULE || type == JSON_MOLECULE || type == CDX_MOLECULE || type == SUBMOLECULE)
+    const static std::unordered_set<int> myTypes = 
+    { 
+        MOLECULE, 
+        QUERY_MOLECULE, 
+        REACTION_MOLECULE, 
+        SCAFFOLD, 
+        RGROUP_FRAGMENT, 
+        RDF_MOLECULE, 
+        SMILES_MOLECULE, 
+        CML_MOLECULE, 
+        JSON_MOLECULE, 
+        CDX_MOLECULE, 
+        SUBMOLECULE
+    };
+
+    if (myTypes.count(type))
         return true;
 
     if (type == ARRAY_ELEMENT)
--- a/api/c/indigo/src/indigo_molecule.cpp
+++ b/api/c/indigo/src/indigo_molecule.cpp
@@ -56,32 +56,24 @@ IndigoMolecule::IndigoMolecule() : IndigoBaseMolecule(MOLECULE)
 
 bool IndigoBaseMolecule::is(const IndigoObject& object)
 {
-    int type = object.type;
-
-    const static int myTypes[] =
+    switch (object.type)
     {
-        MOLECULE,
-        QUERY_MOLECULE,
-        REACTION_MOLECULE,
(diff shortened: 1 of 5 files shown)


api/c/indigo/src/indigo_molecule.cpp
```diff
@@ -58,7 +59,22 @@ bool IndigoBaseMolecule::is(const IndigoObject& object)
...
+        RGROUP_FRAGMENT, 
+        RDF_MOLECULE, 
+        SMILES_MOLECULE, 
+        CML_MOLECULE, 
+        JSON_MOLECULE, 
+        CDX_MOLECULE, 
+        SUBMOLECULE
+    };
+
+    if (myTypes.count(type))
```
mkviatkovskii, line 77: I agree that old "if" was huge and it could be improved, but not with a price of creating a new structure. Maybe we should do it with `switch`:
```c++
switch (object.type)
{
case MOLECULE:
case QUERY_MOLECULE:
case REACTION_MOLECULE:
...
    return true;
case ARRAY_ELEMENT:
    return is(((IndigoArrayElement&)object).get());
default:
    return false;
}
```

api/c/indigo/src/indigo_molecule.cpp
```diff
@@ -36,6 +36,7 @@
 #include "molecule/query_molecule.h"
 #include "molecule/smiles_loader.h"
 #include "molecule/smiles_saver.h"
+#include <unordered_set>
```
mkviatkovskii, line 39: This probably should be deleted (see the comment below), but in general we need to follow a special includes order:

1. special header for current cpp file (if any)

2. standard library headers in alphabetic order

3. other third-party headers in alphabetic order

4. indigo-core headers in alphabetic order

5. current project headers in alphabetic order.

mkviatkovskii (changes requested):
Looks good, thanks. Please merge the master branch and fix couple of cosmetic issues.

api/c/indigo/src/indigo_molecule.cpp
SunFellow: Why not? This is normal trick to decrease complexity. (BTW, switch always significantly increases complexity)
`if` here answers the question "Does type belongs to one of the predefined values?" or "Does it belongs to the set of predefined values?"
Since it is const and static, initialization does not consume time and few memory. The search speed is quick.

api/c/indigo/src/indigo_molecule.cpp
SunFellow: Hmmm, I always see that standard headers (or outside headers in <>) are placed to the end, since (if they are not indirectly included in headers above)  they are very implementation-specific. 
Anyway, they will be reordered according to rules, right?

api/c/indigo/src/indigo_molecule.cpp
SunFellow: OK, how about plain array and plain cycle (sacrificing run-time search speed)?
```
	const static int myTypes[] =
	{
		MOLECULE,
		QUERY_MOLECULE,
		REACTION_MOLECULE,
		SCAFFOLD,
		RGROUP_FRAGMENT,
		RDF_MOLECULE,
		SMILES_MOLECULE,
		CML_MOLECULE,
		JSON_MOLECULE,
		CDX_MOLECULE,
		SUBMOLECULE
	};

	for (const int t : myTypes)
		if (type == t)
			return true;
```

mkviatkovskii: To be honest, I don't get the "switch always significantly increases complexity". It implements exactly the same logic with less amount of code (355 symbols):
```c++
switch (object.type)
{
case MOLECULE:
case QUERY_MOLECULE:
case REACTION_MOLECULE:
case SCAFFOLD:
case RGROUP_FRAGMENT:
case RDF_MOLECULE:
case SMILES_MOLECULE:
case CML_MOLECULE:
case JSON_MOLECULE:
case CDX_MOLECULE:
case SUBMOLECULE:
    return true;
case ARRAY_ELEMENT:
    return is(((IndigoArrayElement&)object).get());
default:
    return false;
}
``` than original (455 symbols) or your (415 symbols). What is more complex in it from your perspective?

api/c/indigo/src/indigo_molecule.cpp
mkviatkovskii: It is being done automatically only by clang-tidy, in case it's correctly set up. We don't use it yet, so for now we should do it manually. clang-format knows nothing about standard library, first-party or third-party code, and only sorts headers that are not splitted with empty line in alphabetical order.

api/c/indigo/src/indigo_molecule.cpp
SunFellow: changed to switch operator

even1024 (approved):

Issue #560, opened 2021-12-17
(R) and (S) stereo labels do not appear for stereo bonds if calculate CIP


AllaZhbanova:
Execute v2/indigo/calculate_cip
struct: 
[stereobonds.txt](https://github.com/epam/Indigo/files/7732647/stereobonds.txt)

options: {smart-layout: true, ignore-stereochemistry-errors: true, mass-skip-error-on-pseudoatoms: false,…}
gross-formula-add-isotopes: true
gross-formula-add-rsites: true
ignore-stereochemistry-errors: true
mass-skip-error-on-pseudoatoms: false
smart-layout: true
output_format: "chemical/x-indigo-ket"

Expected behavior
(R) and (S) stereo labels appear near stereobonds

Actual behavior
(R) and (S) stereo labels do not appear

AllaZhbanova:
Fixed, tested 2.4

Issue #562, opened 2021-12-17
IndigoException: molecule auto loader: Molecule JSON description not found error is shown if click aromatize/dearomatize/Calculate CIP/Calculated Values on empty canvas


AllaZhbanova:
Steps to Reproduce

Execute
/v2/indigo/aromatize
output_format: "chemical/x-indigo-ket"
options:
        aromatize-skip-super-atoms: true
        gross-formula-add-isotopes: true
        gross-formula-add-rsites: true
        ignore-stereochemistry-errors: true
        mass-skip-error-on-pseudoatoms: false
        smart-layout: true
[struct.txt.zip](https://github.com/epam/Indigo/files/7735399/struct.txt.zip)

Expected behavior
Nothing happens.

Actual behavior
Error is shown

*NOTE* the same behavior for dearomatize , layout, calculate_cip, calculate, clean

AllaZhbanova:
fixed, tested 2.4

Issue #548, opened 2021-12-06
Convert method for InChi and InChi Aux formats doesn't accept empty structure


ViktoriiaFedotkina:
Steps to reproduce:
1. Request /v2/indigo/convert with the following options:

options: {smart-layout: true, ignore-stereochemistry-errors: true, mass-skip-error-on-pseudoatoms: false,…}
output_format: "chemical/x-inchi-aux"

or

options: {smart-layout: true, ignore-stereochemistry-errors: true, mass-skip-error-on-pseudoatoms: false,…}
output_format: "chemical/x-inchi"

and struct 
[struct.txt](https://github.com/epam/Indigo/files/7659979/struct.txt)

Expected result:
Receive correct struct  (InChI=1S// string for InChi)

Actual result:
Receiving 400 and error: IndigoException: inchi-wrapper: Indigo-InChI: InChI generation failed: Empty structure. Code: 2.

AllaZhbanova:
Fixed, tested 2.4

Issue #415, opened 2021-06-02
Generation of images works incorrectly


AndreiMazol:
**Steps to reproduce:**
1. Call /render endpoint with the following options (use any struct):
`"options":{"smart-layout":true,"ignore-stereochemistry-errors":true,"mass-skip-error-on-pseudoatoms":false,"gross-formula-add-rsites":true,"render-output-format":"png"}}`
2. Call /render endpoint with the following options (use any struct):
`"options":{"smart-layout":true,"ignore-stereochemistry-errors":true,"mass-skip-error-on-pseudoatoms":false,"gross-formula-add-rsites":true,"render-output-format":"svg"}}`

**Expected results:**
1. Step 1 should return png encoded as base64
2. Step 2 should return svg encoded as base64

**Actual results:**
1. Step 1 returns svg
2. Step 2 return raw svg (without encoding it)

_**NOTES**_
Expected results are based on results of the same method of Standalone mode. I guess API should be aligned.

vitalgerko:
From the side of the API, now everything works correctly, a picture in PNG format is downloaded through Postman. Image download in PNG format does not occur through Ketcher

AllaZhbanova:
The issue is still reproducible
In remote mode create any structure, Click Save As, select Image tab, Select PNG and click "Save As Image"
*Result* Nothing happened
*Expected result* Window is closed, Image is saved in PNG


![2021-12-21_15h34_48](https://user-images.githubusercontent.com/93334478/146931376-805844fa-eaac-4c5b-b5d4-0f98fdd356fc.png)

even1024:
I clearly see that ketcher requests 'image/svg+xml' when png is chosen. 'image/png;base64' must be requested instead.

even1024:
switched to options:render-output-format

AllaZhbanova:
Fixed, tested 2.4

PR #576, opened 2021-12-23
core, wasm: ketcher-related fixes: #560, #548, #415, #562

even1024:


mkviatkovskii (approved):
Look good. Please fix the tests and code formatting and then we'll merge.

PR #577, opened 2021-12-23
feat: dotnet: added automatic binding generation

f1nzer:
Added experimental support for automatic binding generation (.Net) using CppSharp.
Classes with P/Invoke will be generated automatically as a part of the Indigo.Net build process.

That's not a SWIG implementation but at least something 😄 
I would appreciate hearing your thoughts on this!


MysterionRise:
Why it's changes to BingoNosqlLib?

Issue #561, opened 2021-12-17
api: http: implement compounds rendering


mkviatkovskii:
**Motivation**
We want to fully switch to new Indigo FastAPI service  from old one, based on Flask. Most of functionality is already supported, except for compounds rendering.

**ToDo**
Implement `render` endpoint that accepts a compound, output format (png, pdf or svg) and optional key-value list of Indigo renderer options; and returns the rendered image.

See `api/http` and `https://github.com/epam/Indigo/blob/master/utils/indigo-service/service/v2/indigo_api.py` for details.

PR #579, opened 2021-12-26
api: http: implement compounds rendering #561

AATDev21:
Rendering endpoint for png, png_base64, svg+xml, and pdf output formats was implemented.


api/http/tests/test_indigo_http.py
```diff
@@ -1,10 +1,15 @@
...
 import os
 import pathlib
-from typing import Any, List
+import xml.etree.ElementTree as elTree
+from typing import Any, Dict, List
 
+import PyPDF2
 import pytest
 from fastapi.testclient import TestClient
+from PIL import Image
```
mkviatkovskii, line 12: We are using new dependencies here, like `Pillow` or `PyPDF2`, so please add them to `api/http/requirements_dev.txt`.

mkviatkovskii:
Dear @AATDev21, please check the `static_analysis` job result, it's failing due to couple of cosmetic issues.

api/http/indigo_service/jsonapi.py
```diff
@@ -581,3 +582,27 @@ def make_descriptor_response(
...
+
+# Render
+
+
+rendering_formats = {
+    "image/png": "png",
+    "image/png;base64": "png",
+    "image/svg+xml": "svg",
+    "application/pdf": "pdf",
+}
```
mkviatkovskii, line 595: We probably should add `image/svg+xml;base64` as well.

mkviatkovskii: And maybe also some JSON format with returning base64-encoded string. @khyurri, what do you think, could that be useful?

api/http/indigo_service/indigo_http.py
```diff
@@ -206,13 +209,67 @@ def common_bits(
...
+) -> Union[Response, FileResponse]:
+    compound, *_ = service.extract_compounds(compounds(request))
+    output_format = request.data.attributes.outputFormat
+    indigo_renderer = IndigoRenderer(indigo())
+    indigo().setOption(
+        "render-output-format", jsonapi.rendering_formats.get(output_format)
+    )
+    options = request.data.attributes.options
+    for option, value in options.items():
+        indigo().setOption(option, value)
```
mkviatkovskii, line 224: Have you tested, what happens if we also define `render-output-format` in options list? According to the code, it will rewrite the `output_format` defined in request. I believe we need to check if they are equal, and if not, raise an exception, to not misguide the userrs.

api/http/tests/test_indigo_http.py
AATDev21: I noticed a couple of other dependencies and will add them too.

api/http/indigo_service/indigo_http.py
AATDev21: Not yet, and actually I've been thinking about how to escape this situation at all. For example, set `render-output-format ` the obligatory option and the only way to choose the output format. But this is probably worse than the method that you suggested. Thank you, I'll add a new test and check for this case.

api/http/requirements_dev.txt
```diff
@@ -2,12 +2,16 @@
...
 certifi==2021.10.8
 charset-normalizer==2.0.7
+fastapi==0.70.0
 iniconfig==1.1.1
 mypy==0.910
 packaging==21.0
+Pillow==8.4.0
 pluggy==1.0.0
 py==1.10.0
+pydantic==1.8.2
```
mkviatkovskii, line 12: I believe fastapi and pydantic are already included in `requirements.txt`.

mkviatkovskii (approved):

PR #583, opened 2021-12-27
Syntax check findings

SunFellow:
Reviewing changes in PR #570 revealed some points to improve or make the code laconic.

### DIFF
--- a/api/c/indigo/src/indigo_molecule.cpp
+++ b/api/c/indigo/src/indigo_molecule.cpp
@@ -36,6 +36,7 @@
 #include "molecule/query_molecule.h"
 #include "molecule/smiles_loader.h"
 #include "molecule/smiles_saver.h"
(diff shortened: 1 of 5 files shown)


even1024 (approved):

api/c/indigo/src/indigo_molecule.cpp
```diff
@@ -56,15 +56,24 @@ IndigoMolecule::IndigoMolecule() : IndigoBaseMolecule(MOLECULE)
...
+    case CDX_MOLECULE:
+    case SUBMOLECULE:
         return true;
 
-    if (type == ARRAY_ELEMENT)
+    case ARRAY_ELEMENT:
         return is(((IndigoArrayElement&)object).get());
-
+    }
     return false;
```
mkviatkovskii, line 77: Why not using `default:` here?

PR #593, opened 2022-01-05
Restore PosgreSQL 12 support

AlexanderSavelyev:
Background:
After recent changes to TLS the bingo postgres started to fail. The PR fixes the issues by removing all the static dependencies and moving the context into the PostgreSQL userdata object


AlexanderSavelyev:
@mkviatkovskii  all ci builds are restored.
PR contains many files, but the changes only several
1. Move from TLS to local userdata
2. Fix seg fault for exact/gross searches
3. Cosmetic changes in headers (remove using namespace from headers)
Please view at high level if I missed anything

.github/workflows/indigo-ci.yaml
```diff
@@ -995,6 +995,7 @@ jobs:
           case ${{ matrix.postgres_major_version }} in 12) pg_version=12.8;; "11") pg_version=11.13;; 10) pg_version=10.18;; 9.6) pg_version=9.6.23;; esac;
           curl -O https://get.enterprisedb.com/postgresql/postgresql-${pg_version}-1-osx-binaries.zip
           unzip postgresql-${pg_version}-1-osx-binaries.zip
+          mv pgsql/include/zlib.h pgsql/include/zlib.h.bck
```
mkviatkovskii, line 998: Do we still need this?

mkviatkovskii (approved):

.github/workflows/indigo-ci.yaml
AlexanderSavelyev: no, it can be just removed, but following some "devops" mental rules, put move instead...

Issue #594, opened 2022-01-10
Implement clustering procedure for given list of assay values


mkviatkovskii:
**Input:**
* Ordered list of assay values
* Clustering method, default: [SpectralClustering](https://scikit-learn.org/stable/modules/clustering.html#spectral-clustering)
* Clustering params in kwargs format

**Output:** 
* Ordered list of cluster labels

Notes:
* Add code to `api/python/indigo/ml/clustering.py`
* Read about clustering in case of one-dimensional assay values input, probably implement special behavior in this case.

PR #605, opened 2022-01-11
#594 Implement clustering procedure for given list of assay values

suppnova:


api/python/setup.py
```diff
@@ -123,4 +123,5 @@
     },
     download_url="https://pypi.org/project/epam.indigo",
     test_suite="tests",
+    extras_require={"ml": ["scikit-learn", "pytest"]},
```
mkviatkovskii, line 126: We don't need pytest here, since it's test dependency, not runtime.

api/python/indigo/ml/clustering.py
```diff
@@ -0,0 +1,19 @@
+from sklearn.cluster import SpectralClustering
+
+
+def clustering(
+    assay_values: list[list], method=SpectralClustering, **kwargs
+) -> list:
```
mkviatkovskii, line 6: Please specify output list type, I guess it should be `list[int]` in this case.

api/python/indigo/ml/clustering.py
```diff
@@ -0,0 +1,19 @@
+from sklearn.cluster import SpectralClustering
+
+
+def clustering(
+    assay_values: list[list], method=SpectralClustering, **kwargs
```
mkviatkovskii, line 5: Please specify list type, maybe like `list[list[Union[int, float]]]`?

Also `method: Type[ClusterMixing]` seems to be reasonable.

mkviatkovskii (changes requested):
Looks good, just need to fix couple cosmetic issues.

api/python/setup.py
suppnova: Pytest moved to requirements_dev.txt

api/python/indigo/ml/clustering.py
suppnova: Done

api/python/indigo/ml/clustering.py
suppnova: Done. 
But ClusterMixing caused some IDE highlights. Still thinking about collecting all valid clustering methods into a list for type checking

mkviatkovskii (approved):

Issue #597, opened 2022-01-10
Add notebook for predicting assay values using Indigo fingerprints


mkviatkovskii:
**ToDo**
Prepare and place a notebook in api/python/notebooks/predict.ipynb
* Read input CSV file, with given column name for SMILES and column name for assay values.
* Prepare train and test datasets using shuffled data.
* Draw distribution plot for assay values.
* Calculate fingerprints for SMILES using Indigo.
* Fit and predict for two models: linear on your choice (SVM, RF) and deep learning mode on your choice. Print `r^2` value for predicting values for test dataset.
* Draw an interactive scatter plots using [Bokeh](https://docs.bokeh.org/en/2.4.1/docs/user_guide/jupyter.html). Use hover tool to show molecule images, actual and predicted value. Following plots should present:
  * Actual vs Predicted
  * Actual vs Residual (actual - predicted)

PR #608, opened 2022-01-13
 Add notebook for predicting assay values using Indigo fingerprints #597 

alkornev:


mkviatkovskii (changes requested):
Please:

1. Rename TARGET to ASSAY
2. Add simple matplotlib plots for AvP and AvR.
3. Consider moving large code pieces into python modules to keep notebook clean and simple.

mkviatkovskii (approved):

Issue #596, opened 2022-01-10
Add notebook for clustering and manifolding for given set of molecules and assay values with proper visualization


mkviatkovskii:
Is blocked by #594, #595.

**ToDo**
1. Read input CSV file, with given column name for SMILES and column name(s) for assay values.
2. Clusterize assay values using method from #594.
3. Calculate fingerprints for SMILES using Indigo and then calculate 2D coordinates using method from #595.
4. Draw an interactive scatter plot using [Bokeh](https://docs.bokeh.org/en/2.4.1/docs/user_guide/jupyter.html), use 2D coordinates, cluster labels. Add hover tool with rendered compound image.

PR #610, opened 2022-01-17
#596 Add notebook for clustering and manifolding for given set of molecules

suppnova:


mkviatkovskii (changes requested):
Please:
1. Rename comments assay_1 and assay_2 to "logP" and name of assay correspondingly.
2. Rename "simple plot", since it's quite specific and accepts "cluster" as input,
3. Consider modifying fingerprints parameters to achieve better splits between low and high values on manifold plots (May be done in standalone pull request).
4. Merge fresh master.

mkviatkovskii (approved):

Issue #603, opened 2022-01-10
Move Bingo Postgres tests to GitHub and add them to CI


mkviatkovskii:
We have existing Bingo Tests in our old repository. We need to:
1.  Check locally how do they work with fresh Bingo and Postgres.
2. Move the code to GitHub.
3. Add README.md and describe how to start tests locally.
4. Add Dockerfile for Bingo-Postgres and build/push it in CI.
5. Add bingo-tests to CI after Bingo-Postgres build step, use Docker image for that.

PR #611, opened 2022-01-18
Move Bingo postgres tests to GitHub #603

AATDev21:
Moved Bingo tests to GitHub, added Dockerfile for Bingo-Postgres image and added tests to CI.


core/indigo-core/molecule/src/cml_saver.cpp
```diff
@@ -361,6 +361,10 @@ void CmlSaver::_addMoleculeElement(XMLElement* elem, BaseMolecule& mol, bool que
                 }
                 else
                 {
+                    if (pos.x == 0)
+                        pos.x = 0;
+                    if (pos.y == 0)
+                        pos.y = 0;
```
SunFellow, line 367: If a value already is zero, set to zero.
Strange logic block that does nothing. 
Looking to above clause, it can be removed.

AATDev21: It was desperate attemp to eliminate negative zeros from some coordinates in xml, which have started to appear after updating TinyXML with TinyXML2. It is an example of xml, produced by TinyXML2:

image

SunFellow: C++ float numbers really have some peculiarities regarding `-0f`. 
Thinking about coordinates, they can hardly be essential, can't they?
Anyway, if the problem comes from reading XML, the correction should be made in the corresponding layer.
`MoleculeCdxmlLoader` ?

SunFellow:
`.github/workflows/indigo-ci.yaml` was chaning many times back and forth and merged changes from master.
It must be clean from intermediate changes. 
If we want to add a step for bingo tests, it should be "one-line" addition calling the corresponding script that can be run independently of GiHub.

core/indigo-core/molecule/src/cml_saver.cpp
SunFellow: So I'd propose the code like this:
```
    if (pos.x == -0.f)
        pos= 0.f;
```
Or even better (quasi quantization around zero):
```
    if (fabs(pos.x) < FLT_MIN)
        pos= 0.f;
```

AATDev21: Thank you, @SunFellow! And extend it to the rest coordinates just in case, right?

SunFellow: Yes, of course.

AlexanderSavelyev (approved):
All changes were reviewed. The tests are ready to be merged

SunFellow (approved):

MysterionRise (approved):

Issue #525, opened 2021-11-14
Convert error when User tryes to save structure in Daylight SMARTS format


AllaZhbanova:
***Steps to reproduce:***

1. call indigo/convert
options: {smart-layout: true, ignore-stereochemistry-errors: true, mass-skip-error-on-pseudoatoms: false,…}
output_format: "chemical/x-daylight-smarts"

struct: 
[DS.zip](https://github.com/epam/Indigo/files/7563041/DS.zip)

**Expected Result:**
SMARTS string is generated.

***Actual Result:***
{"error":"IndigoException: SMILES saver: SMARTS format availble for query only"}

AllaZhbanova:
The issue is still reproducible
![2021-12-14_15h19_07](https://user-images.githubusercontent.com/93334478/145997234-9899215b-f0a9-4992-bc96-e4788fa37fc5.png)

Issue #478, opened 2021-09-06
Error message from indigo when trying open SMART format file with any Atom Generic  and Pol


AleksandraIvleva:
**Steps to Reproduce**

1. Execute /v2/indigo/convert with the following parameters (Pol in structure):
{
  "struct": "\n  Ketcher  9 62112292D 1   1.00000     0.00000     0\n\n  6  5  0  0  0  0            999 V2000\n    4.7250   -6.0750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n    5.2250   -6.9410    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n    6.2250   -6.9410    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n    6.7250   -7.8071    0.0000 Pol 0  0  0  0  0  0  0  0  0  0  0  0\n    7.7250   -7.8071    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n    8.2250   -8.6731    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n  1  2  1  0     0  0\n  2  3  1  0     0  0\n  3  4  1  0     0  0\n  4  5  1  0     0  0\n  5  6  1  0     0  0\nM  END\n",
  "output_format": "chemical/x-daylight-smarts",
  "options": {
    "smart-layout": true,
    "ignore-stereochemistry-errors": true,
    "mass-skip-error-on-pseudoatoms": false,
    "gross-formula-add-rsites": true,
    "gross-formula-add-isotopes": true
  }
}
Error message. Response:
"IndigoException: SMILES saver: SMARTS format availble for query only"

2. Execute /v2/indigo/convert with the following parameters (the same structure as in Step1 + A):
{
[...]

PR #617, opened 2022-01-22
core: SMARTS/SMILES loader and saver bugfixes #525, #478

even1024:


api/c/indigo/src/indigo_savers.cpp
```diff
@@ -198,11 +199,23 @@ void IndigoSmilesSaver::generateSmarts(IndigoObject& obj, Array<char>& out_buffe
...
         else
-            saver.saveMolecule(mol.asMolecule());
+        {
+            Array<char> mol_out_buffer;
+            ArrayOutput mol_output(mol_out_buffer);
+            MolfileSaver saver_tmp(mol_output);
+            saver_tmp.saveMolecule(mol.asMolecule());
+            mol_out_buffer.push(0);
+            BufferScanner sc(mol_out_buffer);
+            MolfileLoader loader_tmp(sc);
```
mkviatkovskii, line 214: Maybe it's more correct to use SMILES loader/saver in this case?

mkviatkovskii (approved):

api/c/indigo/src/indigo_savers.cpp
even1024: Unfortunately SMILES parsers are buggy. We can't rely upon on it right now.

PR #620, opened 2022-01-23
bingo-postgres: fix bug #604 with error logs

### LINKED ISSUE #604: bingo-postgres: error when executing ImportSDF
Could be reproduced on fresh master (commit g47a793661) at least on PostgreSQL 14 and macOS:
```sql
create table test(id serial, molfile varchar);
select bingo.ImportSDF('test', 'molfile', '', '/tmp/indigo/data/molecules/basic/sugars.sdf');
```
Output:
```
[2022-01-10 17:54:46] [XX000] ERROR: errstart was not called
```
The problem is gone if we change `molfile` column type to `text`. Still it is very unclear error message that gives zero useful information about the source of the problem.

AlexanderSavelyev:
FIxes #604

### DIFF
--- a/bingo/postgres/src/pg_common/bingo_pg_common.h
+++ b/bingo/postgres/src/pg_common/bingo_pg_common.h
@@ -432,13 +432,13 @@ private:
     catch (indigo::Exception & e)                                                                                                                              \
     {                                                                                                                                                          \
         pg_raise_error = true;                                                                                                                                 \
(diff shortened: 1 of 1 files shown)


mkviatkovskii (approved):

AATDev21 (approved):
"errstart was not called" vanished with postgresql versions 13 and 14

Issue #602, opened 2022-01-10
Implement simple minimal mol2graph procedure and application using it


mkviatkovskii:
**ToDo**
We need to Implement our own minimal mol2graph procedure using Torch, DGL and Indigo. There is an existing implementation that uses RDKit, we need to reimplement it with Indigo
1. Add simple atom and bonds featurizers (atom number and bond order). See https://github.com/awslabs/dgl-lifesci/blob/master/python/dgllife/utils/featurizers.py for details
2. Add simple mol2graph using featurizers, see https://github.com/awslabs/dgl-lifesci/blob/master/python/dgllife/utils/mol_to_graph.py for details
3. Add simple code (in notebook and/or CLI form) to use it for predicting molecular properties, see https://github.com/awslabs/dgl-lifesci/tree/master/examples/property_prediction/MTL for details.

PR #621, opened 2022-01-24
#602 Implement simple minimal mol2graph procedure and application using it 

DianaPrs:


api/python/properties_prediction_d/config.py
```diff
@@ -0,0 +1,24 @@
+"""Config file"""
+
+
+# Dataset
+
+file_name = "Adrenergic_dataset.csv"
+target = "logP"  # "AdrA1A_PCHEMBL_VALUE"
+smiles = "Structure"
```
mkviatkovskii, line 8: I think, this should not be predefined. Better to set it as input parameters in the notebook or CLI.

api/python/properties_prediction_d/datasets.py
```diff
@@ -0,0 +1,53 @@
+import config
```
mkviatkovskii, line 1: We should add these files to `api/python/indigo/ml`, and call it like `mpp` (molecular properties prediction).

api/python/properties_prediction_d/datasets.py
```diff
@@ -0,0 +1,53 @@
...
+import pandas as pd  # type: ignore
+from dgl.data import DGLDataset  # type: ignore
+from dgl.data.utils import split_dataset  # type: ignore
+from dgl.dataloading import GraphDataLoader  # type: ignore
+from mol_to_graph import smiles_to_graph
```
mkviatkovskii, line 6: We need to import first-party modules (e.g. our own code) after third-party imports, and use this approach:

```python
from indigo.ml.mpp.mol_to_graph import smiles_to_graph
```

api/python/properties_prediction_d/mol_to_graph.py
```diff
@@ -0,0 +1,63 @@
...
+    if node_featurizer is not None:
+        g.ndata.update(node_featurizer(mol))
+
+    if edge_featurizer is not None:
+        g.edata.update(edge_featurizer(mol))
+
+    return g
+
+
+def smiles_to_graph(smiles):
```
mkviatkovskii, line 36: Technically it is `mol_to_graph`, since `indigo.loadMolecule` automatically guesses input format and can load SMILES, InChI, Molfile, CML and couple of other formats.

mkviatkovskii (changes requested):
Looks fine. Please take a look at the comments and also consider adding type hints where possible.

Also please delete extra notebook for fingerprint-based prediction, we don't need it in this request.

mkviatkovskii (approved):

Issue #642, opened 2022-02-03
Bingo Elastic in Python 1.6.1 doesn't contain model classes in published package


MysterionRise:
1.6.0 version - with model dir and classes inside

<img width="638" alt="Screenshot 2022-02-03 at 22 25 01" src="https://user-images.githubusercontent.com/956191/152439423-9c570548-c66e-4f75-8196-73e90b5ccbe2.png">

1.6.1 version - no such files, hence it's not working

<img width="565" alt="Screenshot 2022-02-03 at 22 25 51" src="https://user-images.githubusercontent.com/956191/152439533-98916aa2-d699-4d48-9131-7eef71a4309d.png">

@khyurri , any ideas? I suspect migration from poverty to pip...

MysterionRise:
@mkviatkovskii when could we release this bug fix? Currently recent version is broken

MysterionRise:
Fixed in 1.7.0 release
@mkviatkovskii  not included in the release notes

PR #643, opened 2022-02-04
bingo-elastic-python: update setup.py to include model subpackage

MysterionRise:
fixing #642


MysterionRise:
@mkviatkovskii @khyurri please take a look, if all is fine, we also need to update current PyPi package, since latest is really not-working now

mkviatkovskii (approved):

PR #652, opened 2022-02-09
#613 Implement atom orbital hybridization computation

suppnova:


api/python/indigo/hybridization.py
```diff
@@ -0,0 +1,222 @@
...
+    if n_bonds == 0:
+        return "s"
+
+    if atom.symbol() in ["C", "N", "O", "P", "S", "B"] and in_aromatic_ring(
+        atom
+    ):
+        return "sp2"
+
+    if atom.atomicNumber() == 6:
+        return cardon_hybridization(atom)
```
mkviatkovskii, line 206: carBon

api/python/indigo/hybridization.py
```diff
@@ -0,0 +1,222 @@
...
+        "sp2", "sp3", "sp3d", "sp3d2", "sp3d3", "sp3d4" for hybridized.
+    """
+    # if the atomic number is undefined or ambiguous
+    if atom.atomicNumber() == 0:
+        raise IndigoException("Atomic number is undefined or ambiguous")
+
+    # don't bother with the lantanoids and beyond
+    elif atom.atomicNumber() >= 57:
+        raise IndigoException(
+            "Hybridization is not calculated for atomic number >= 57"
```
mkviatkovskii, line 190: "Hybridization calculation is not implemented for atomic numbers >= 57"

api/python/indigo/hybridization.py
```diff
@@ -0,0 +1,222 @@
...
+    if neighbors == 6:
+        return "sp3d2"
+    if neighbors == 7:
+        return "sp3d3"
+    if neighbors == 8:
+        return "sp3d4"
+    return None
+
+
+def in_aromatic_ring(atom):
```
mkviatkovskii, line 163: Please add type hints

mkviatkovskii (changes requested):
Looks impressive, thanks!

Left some cosmetic related comments, and also just a thought — maybe we also need to add an Enum for hybridization types and operate with it. And add two methods:

getHybridization -> Enum
getHybridizationStr -> str

This could simplify using hybridization in ML featurization and also avoid possible mistakes with operating with raw strings. What do you think?

api/python/indigo/hybridization.py
```diff
@@ -0,0 +1,225 @@
...
+    "I": 7,
+    "Xe": 8,
+    "Cs": 1,
+    "Ba": 2,
+    "La": 3,  # 6s2 5d1
+    "Ce": 4,  # 6s2 4f2 (!!!) Atomic num = 58, lantanoids
+}
+
+
+class EnumHybridizations(Enum):
```
mkviatkovskii, line 72: Let's rename it to HybridizationType.

suppnova: Done

mkviatkovskii (approved):

mkviatkovskii (approved):

Issue #623, opened 2022-01-25
Implement calculation of logP for molecule


mkviatkovskii:
1. Read article: https://www.researchgate.net/profile/Gordon-Crippen/publication/220524632_Prediction_of_Physicochemical_Parameters_by_Atomic_Contributions/links/5d6c271aa6fdcc547d720a35/Prediction-of-Physicochemical-Parameters-by-Atomic-Contributions.pdf
2. Implement calculation of logP based on method described in the article
3. Add tests for some compounds from Adrenergic dataset to check is our logP value is equal to the value from ChemBL.

PR #658, opened 2022-02-16
#623 Implement calculation of logP and MR for molecule

DianaPrs:


api/python/indigo/__init__.py
```diff
@@ -37,6 +37,8 @@
     sizeof,
 )
 
+from logP import get_logp, get_mr  # type: ignore
```
mkviatkovskii, line 40: Please rename logP module to "logp.py", it's agaist PEP8.

api/python/indigo/__init__.py
```diff
@@ -2541,6 +2543,22 @@ def iterateBonds(self):
...
 
+    def logP(self, csv_file=None) -> float:
+        """Molecule method returns calculated logP value
+
+        Returns:
+            float: calculated logP value of the molecule
+        """
+        return get_logp(self, csv_file=None)
+
+    def MR(self, csv_file=None) -> float:
```
mkviatkovskii, line 2554: Thanks for adding this too.

Let's rename it to "molarRefractivity", "MR" sounds unclear.

api/python/indigo/logP.py
```diff
@@ -0,0 +1,160 @@
...
+from collections import Counter, defaultdict
+from typing import Dict, List
+
+from smarts_table import TABLE  # type: ignore
+
+from indigo import Indigo, IndigoObject  # type: ignore
+
+indigo = Indigo()
+indigo.setOption("ignore-stereochemistry-errors", True)
+indigo.setOption("ignore-bad-valence", True)
```
mkviatkovskii, line 16: I'm afraid it won't work like that. Let's discuss it in private.

api/python/indigo/logP.py
```diff
@@ -0,0 +1,160 @@
...
+        float: MR value
+    """
+    values = []
+    mr_contributions = get_mr_contributions_dict(table)
+    for t, n in types_counter.items():
+        values.append(mr_contributions[t] * n)
+    return sum(values)
+
+
+def count_logp(types_counter: Dict[str, int], table: List[str]) -> float:
```
mkviatkovskii, line 81: calculate_logp should be better

api/python/indigo/logP.py
```diff
@@ -0,0 +1,160 @@
...
+        dictionary of MR contributions to atom types
+    """
+    contrib_dict = {}
+    for string in types_smarts:
+        row = string.split()
+        contrib_dict[row[0]] = float(row[3])
+    return contrib_dict
+
+
+def count_mr(types_counter: Dict[str, int], table: List[str]) -> float:
```
mkviatkovskii, line 65: calculate_mr should be better

mkviatkovskii (changes requested):
Looks fine, except for couple cosmetic issues and lack of tests. Also we need to discuss one technical problem.

mkviatkovskii (approved):

Issue #522, opened 2021-11-11
core: replace `Obj` with standard smart pointer


mkviatkovskii:
**Motivation**
Currently Indigo uses a strange template class `Obj<T>` defined in `core/indigo-core/common/base_cpp/obj.h`. It allocates memory for `T` and can create an instance of `T` there, give access to it using `->` operator and clear memory.
So it looks much like `std::unique_ptr`, at least at the first sight. And it contains several template `create` methods instead of variadic template `make_unique`.

**ToDo**
1. Find all places in code that use `Obj<T>`.
2. Check if they could be replaced with `T` (no null state, no replacing object in container, `T` has default constructor etc) and replace it with `T` if possible.
3. If not possible, try to replace it with `std::unique_ptr<T>`.
4. In some cases this class returns a raw pointer to contained instance of `T`. Check if it makes sense to use a `std::shared_ptr<T>` instead to make it more convenient and safe.

AlexeyGirin:
Verified.
![Image](https://github.com/user-attachments/assets/2d0362e6-88b4-4d9c-b8fa-6cb7414a4b58)

PR #670, opened 2022-02-20
core: replace Obj with standard smart pointer #522

malysh123:
Fixes [#522](https://github.com/epam/Indigo/issues/522)

### DIFF
--- a/core/indigo-core/molecule/src/inchi_wrapper.cpp
+++ b/core/indigo-core/molecule/src/inchi_wrapper.cpp
@@ -602,7 +602,7 @@ void InchiWrapper::saveMoleculeIntoInchi(Molecule& mol, Array<char>& inchi)
             break;
         }
 
-    Molecule* target = &mol;
+    std::unique_ptr<Molecule> target(&mol);
 
     if (has_aromatic)
     {
@@ -623,7 +623,7 @@ void InchiWrapper::saveMoleculeIntoInchi(Molecule& mol, Array<char>& inchi)
         catch (DearomatizationException&)
         {
         }
-        target = dearom.get();
+        target = std::move(dearom);
     }
     generateInchiInput(*target, input, atoms, stereo);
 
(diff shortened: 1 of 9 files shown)


core/indigo-core/molecule/src/inchi_wrapper.cpp
```diff
@@ -603,10 +603,11 @@ void InchiWrapper::saveMoleculeIntoInchi(Molecule& mol, Array<char>& inchi)
         }
 
     Molecule* target = &mol;
-    Obj<Molecule> dearom;
+    std::shared_ptr<Molecule> dearom;
+
     if (has_aromatic)
     {
-        dearom.create();
+        dearom = std::shared_ptr<Molecule>();
```
mkviatkovskii, line 610: Looks like we can use `unique_ptr` here, lifetime of the object is straight and limited. Moreover, we can move declaration of `dearom` inside `if` clause.

core/indigo-core/molecule/src/molecule_fingerprint.cpp
```diff
@@ -600,12 +600,12 @@ void MoleculeFingerprintBuilder::_handleSubgraph(Graph& graph, const Array<int>&
...
-    Obj<TautomerSuperStructure> tau_super_structure;
+    std::shared_ptr<TautomerSuperStructure> tau_super_structure;
     BaseMolecule* mol_for_enumeration = &mol;
 
     if (!query && _parameters.tau_qwords > 0 && !skip_tau)
     {
-        tau_super_structure.create(mol.asMolecule());
+        tau_super_structure = std::make_shared<TautomerSuperStructure>(mol.asMolecule());
 
         _tau_super_structure = tau_super_structure.get();
```
mkviatkovskii, line 610: I believe it's better to use `unique_ptr` here and even change type of `_tau_super_structure` to `std::unique_ptr`.

malysh123: But we have assignment operator here.

mkviatkovskii: Let's take a look at this and discuss together.

PR #698, opened 2022-03-30
9.6 postgres sql support removal from linux x86 build

even1024:
Remove build of [linux_x86_64 (9.6) build. Postgres sql 9.6 not available.

### DIFF
--- a/.github/workflows/indigo-ci.yaml
+++ b/.github/workflows/indigo-ci.yaml
@@ -929,7 +929,7 @@ jobs:
         run: git fetch --tags -f
       - name: Install postgres server headers
         run: |
-          case ${{ matrix.postgres_major_version }} in 12) pg_version=12;; 11) pg_version=11;; 10) pg_version=10;; esac;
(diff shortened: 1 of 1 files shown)


SunFellow (commented):
1. I propose to confirm the change with Aleksander Savelyev.
2. Probably, we should remove support Postgres 9.6 on other platforms as well.
3. Probably, this change should be accompanied with removing the "promise" about testing on versions 9.x / 9.6 from corresponding files, e.g. `bingo/postgres/README.md`

.github/workflows/indigo-ci.yaml
```diff
@@ -929,7 +929,7 @@ jobs:
         run: git fetch --tags -f
       - name: Install postgres server headers
         run: |
-          case ${{ matrix.postgres_major_version }} in 12) pg_version=12;; 11) pg_version=11;; 10) pg_version=10;; 9.6) pg_version=96;; esac;
```
SunFellow, line 932: This line can be left unchanged. Dropping 9.6 is provided by line#919.

SunFellow (approved):

.github/workflows/indigo-ci.yaml
even1024: fixed

Issue #713, opened 2022-04-20
api: python: implement salt stripping method


AATDev21:
**Description**
We need to extend standardization procedure by adding a method, which removes all inorganic fragments from multifragment molecules. The example is shown on the picture below:

<img width="611" alt="image" src="https://user-images.githubusercontent.com/78915761/164222891-f9a2c8a7-0743-428f-80a5-53be30bba06a.png">

**ToDo**
Implement a method for `IndigoObject` that accepts a molecule and optional flag `inplace`, which specifies a mode for the molecule modification.

SunFellow:
@AATDev21 , @mkviatkovskii ,
Question: Do we need to add the same method to other languages libs: .NET and Java?

mkviatkovskii:
> @AATDev21 , @mkviatkovskii ,
> 
> Question: Do we need to add the same method to other languages libs: .NET and Java?
> 

We definitely would like to have it, but someone should reimplement it in C++ first.

PR #716, opened 2022-04-20
#713 api: python: implement salt stripping method

AATDev21:
`saltStrip()` method was implemented for removing from the molecule all disconnected inorganic components.

### DIFF
--- a/api/python/indigo/__init__.py
+++ b/api/python/indigo/__init__.py
@@ -1266,6 +1266,41 @@ class IndigoObject(object):
                     return True
         return False
 
+    def stripSalt(self, inplace=False):
+        """Molecule method strips all inorganic components.
+
+        Args:
+            inplace(bool): if False - returns the copy of the molecule,
+                           without inorganic components, if True - strips
+                           inorganic components from the molecule itself.
+
+        Returns:
+            IndigoObject or None: new molecule without inorganic components
+                                  if inplace=False, None otherwise.
+        """
+        self.dispatcher._setSessionId()
+        salts_atoms = []
+        idx = 0
+        for target_fragment in self.iterateComponents():
+            target_fragment = target_fragment.clone()
+            n_atoms = target_fragment.countAtoms()
+
+            for salt in SALTS:
+                query_salt = self.dispatcher.loadQueryMolecule(salt)
+                matcher = self.dispatcher.substructureMatcher(target_fragment)
+                if matcher.match(query_salt):
+                    salt_position = [i for i in range(idx, idx + n_atoms)]
+                    salts_atoms.extend(salt_position)
+            idx += n_atoms
+
+        if not inplace:
+            saltless_fragment = self.clone()
+            saltless_fragment.removeAtoms(salts_atoms)
+            return saltless_fragment
+        else:
+            self.removeAtoms(salts_atoms)
+            return None
+
     def countHydrogens(self):
         """Atom or Molecule method returns the number of hydrogens
 
--- a/api/python/indigo/__init__.py
+++ b/api/python/indigo/__init__.py
@@ -1275,8 +1275,10 @@ class IndigoObject(object):
                            inorganic components from the molecule itself.
 
         Returns:
-            IndigoObject: new molecule without inorganic components
-                          if inplace=False, initial molecule otherwise.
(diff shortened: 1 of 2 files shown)


api/python/indigo/__init__.py
```diff
@@ -1266,6 +1266,41 @@ def checkSalt(self):
...
         return False
 
+    def stripSalt(self, inplace=False):
+        """Molecule method strips all inorganic components.
+
+        Args:
+            inplace(bool): if False - returns the copy of the molecule,
+                           without inorganic components, if True - strips
+                           inorganic components from the molecule itself.
+
```
SunFellow, line 1276: IMHO, it's better to design method as thin as possible: to solve just a single task.
This parameter overloads a single method with two tasks: clone and strip, while clone is already implemented. 
My proposal: leave only in-place mode. If the caller needs to operate on a copy, it can create it first.

SunFellow: Other hand, might be useful "read-only" Boolean method that checks if there is any atom to be stripped. Than the caller can call it first and skip unnecessary clone, since the resulting object will be the same.

api/python/tests/test_indigo.py
```diff
@@ -117,72 +96,117 @@ def test_check_complex_ion(self) -> None:
...
         )
+
+        # Complex organic salt
+        m5 = self.indigo.loadMolecule(
+            "C(C(C(C(C(C(=O)O)O)O)O)O)O.C(C(C(C(C(C(=O)O)O)O)O)O)O.[Fe]"
+        )
+
+        # No organic components
+        m6 = self.indigo.loadMolecule("[NH4+].[O-]P(=O)([O-])[O-].[Fe+2]")
+
```
SunFellow, line 185: If these `m1`... `m6` are independent, then it's better to make one (independent) test per molecule.
This refers to other tests as well, but they did exist before.

api/python/indigo/__init__.py
mkviatkovskii: `inplace: bool = False` is a standard pattern in Python code, so I suggest continue using it.

api/python/indigo/__init__.py
```diff
@@ -1266,6 +1266,41 @@ def checkSalt(self):
...
+                    salts_atoms.extend(salt_position)
+            idx += n_atoms
+
+        if not inplace:
+            saltless_fragment = self.clone()
+            saltless_fragment.removeAtoms(salts_atoms)
+            return saltless_fragment
+        else:
+            self.removeAtoms(salts_atoms)
+            return None
```
mkviatkovskii, line 1302: This is a bad behavior, we violate "single return type" rule. We should return the same type in any case. I think, `return self` is correct.

mkviatkovskii (changes requested):
Please take a look at the comments and fix the request according to them.

api/python/tests/test_indigo.py
AATDev21: All tests were fixed, thank you!

api/python/indigo/__init__.py
AATDev21: Fixed and remembered for the future 🤔

mkviatkovskii (approved):

api/python/indigo/__init__.py
```diff
@@ -1275,8 +1275,8 @@ def stripSalt(self, inplace=False):
                            inorganic components from the molecule itself.
 
         Returns:
-            IndigoObject or None: new molecule without inorganic components
-                                  if inplace=False, None otherwise.
+            IndigoObject: new molecule without inorganic components
+                          if inplace=False, initial molecule otherwise.
```
SunFellow, line 1279: IMHO, something unclear is here:
`if inplace=False, initial molecule otherwise.`

"initial molecule" can be understood as initial molecule without any changes. I can only guess that it will be so if noting to remove...
Please give more detailed description.

AATDev21: Yes, it's indeed wrong. Fixed it too!

SunFellow (approved):

IuriiPuzanov (approved):

mkviatkovskii (approved):

PR #741, opened 2022-05-26
ci: fix npm versioning according to semver

mkviatkovskii:


SunFellow:
I seem the approach to format a version from tags in the repo is wrong. Just the current state of the checked-out commit files matters. 

So since `package.json` in the sources already contains a version, why we can't use it?
If the version of the npm package is different entity then the version in the source and must be unique, then let's think about options:
* remove `version` field from `package.json` or leave a template; so building will set only build number (e.g. the fourth part of the combined version).
* completely remove/rename `package.json` from the sources as confusing (duplicating); instead put a template file with **input** information to be taken for the resulting package.json.

PR #753, opened 2022-06-02
Replace Obj<T> to std::unique_ptr<T>

SunFellow:
1. Get rid of class `Obj<T>` and containing file.
2. Replace all instances of template `Obj<T>` to `std::unique_ptr<T>`
3. Use `std::make_unique<T>(...)` instead `.create(...)`
4. Fix absent methods and operators.


MysterionRise:
Why the builds are failing?

SunFellow:
> Why the builds are failing?
It looks like that dynamic memory reallocations by `std::unique_ptr<>` revealed errors in managing objects. Previously `Obj<>` reused the same memory buffer. Have to investigate.

mkviatkovskii:
Won't work until we completely get rid of own-written memcpy-based containers. Let's wait until we rewrite them.

PR #755, opened 2022-06-04
metadata refactoring

even1024:


core/indigo-core/molecule/base_molecule.h
```diff
@@ -99,33 +99,50 @@ namespace indigo
     class Molecule;
     class QueryMolecule;
 
-    class MetaObjectsInterface
+    class DLLEXPORT MetaData
```
mkviatkovskii, line 102: Let's move this class to standalone file, and also let's rename it, Metadata is too generic. At least to BaseMoleculeMetadata, if you don't have better name.

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -4325,3 +4313,81 @@ void BaseMolecule::getBoundingBox(Rect2f& bbox) const
     }
     bbox = Rect2f(a, b);
 }
+
+void MetaData::addMetaObject(MetaObject* pobj)
```
mkviatkovskii, line 4317: Why do we keep using raw pointers?

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -4325,3 +4313,81 @@ void BaseMolecule::getBoundingBox(Rect2f& bbox) const
...
+    }
+}
+
+void MetaData::cloneMetaData(const MetaData& other)
+{
+    resetMetaData();
+    const auto& meta = other.metaData();
+    for (int i = 0; i < meta.size(); i++)
+        addMetaObject(meta[i]->clone());
+}
```
mkviatkovskii, line 4345: Why do we keep using this clone methods instead of copy constructors?

core/indigo-core/molecule/base_molecule.h
even1024: renamed to MetaDataStorage and moved to metadata_storage.h and metadata_storage.cpp

core/indigo-core/molecule/src/base_molecule.cpp
even1024: Unfortunately we still can't use std classes as members in Indigo.

core/indigo-core/molecule/src/base_molecule.cpp
even1024: Unfortunately we still can't use std classes as members in Indigo.

core/indigo-core/molecule/metadata_storage.h
```diff
@@ -0,0 +1,71 @@
...
+        {
+        }
+
+        void addMetaObject(MetaObject* pobj);
+
+        void resetMetaData()
+        {
+            _meta_data.clear();
+            _plus_indexes.clear();
+            _arrow_indexes.clear();
```
mkviatkovskii, line 50: I believe metadata for molecule and reaction should be different — there should not be any `plus` or `arrow` in molecule metadata. Perhaps we should think about inheritance here?

core/indigo-core/molecule/src/base_molecule.cpp
mkviatkovskii: I understand the reason why we can't use `std`, but why we can't use copy constructors instead of `clone` methods?

core/indigo-core/molecule/metadata_storage.h
even1024: Metadata shouldn't be different for molecules and reactions as far as plus or arrow doesn't always mean "reaction".
It should be possible to load arrows and pluses with molecules.

core/indigo-core/molecule/metadata_storage.h
```diff
@@ -0,0 +1,71 @@
...
+        }
+        uint32_t _class_id;
+        virtual MetaObject* clone() const = 0;
+        virtual ~MetaObject(){};
+    };
+
+    class MetaDataStorage
+    {
+    public:
+        void cloneMetaData(const MetaDataStorage& other);
```
SunFellow, line 38: Proposing shorter name: `clone(...)`

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -4325,3 +4313,81 @@ void BaseMolecule::getBoundingBox(Rect2f& bbox) const
...
+        return *_meta_data[_simple_object_indexes[index]];
+        break;
+    case KETReactionPlus::CID:
+        return *_meta_data[_plus_indexes[index]];
+        break;
+    case KETReactionArrow::CID:
+        return *_meta_data[_arrow_indexes[index]];
+        break;
+    default:
+        break;
```
SunFellow, line 4361: Should return some object's reference or throw an exception.

core/indigo-core/molecule/base_molecule.h
```diff
@@ -98,34 +99,16 @@ namespace indigo
...
-    class DLLEXPORT BaseMolecule : public Graph, public MetaObjectsInterface
+    class DLLEXPORT BaseMolecule : public Graph
     {
     public:
         typedef RedBlackMap<int, int> Mapping;
 
         BaseMolecule();
         ~BaseMolecule() override;
-
+        MetaDataStorage& meta();
```
SunFellow, line 111: Bad practice: return a non-const reference of internal data.

core/indigo-core/molecule/src/base_molecule.cpp
```diff
@@ -4325,3 +4313,81 @@ void BaseMolecule::getBoundingBox(Rect2f& bbox) const
...
+void MetaDataStorage::resetReactionData()
+{
+    _plus_indexes.clear();
+    _arrow_indexes.clear();
+    for (int i = _meta_data.size() - 1; i >= 0; i--)
+    {
+        if (_meta_data[i]->_class_id == KETReactionArrow::CID || _meta_data[i]->_class_id == KETReactionPlus::CID)
+            _meta_data.remove(i);
+    }
+}
```
SunFellow, line 4393: All these method seems of class `MetaDataStorage`. Why to repeat them here?

core/indigo-core/reaction/base_reaction.h
```diff
@@ -92,6 +92,8 @@ namespace indigo
         BaseReaction();
         virtual ~BaseReaction();
 
+        MetaDataStorage& meta();
```
SunFellow, line 95: Bad practice: return non-const reference/pointer to internal data.

core/indigo-core/molecule/metadata_storage.h
even1024: Fixed

core/indigo-core/molecule/src/base_molecule.cpp
even1024: Fixed

core/indigo-core/molecule/src/base_molecule.cpp
even1024: Fixed

mkviatkovskii (approved):

Issue #764, opened 2022-06-17
'Not list' information is lost when structure is saved as 'Daylight SMARTS'


AndreiMazol:
**Steps to Reproduce**
1. create any structure (MOL)  with 'Not List' element (the list should contain 2or more elements)
2. convert created structure to SMARTS
3. convert SMARTS to mol

**Expected behavior**
structure1 (step1) equals to structture2 (step3)

**Actual behavior**
'Not LIst' information is lost.

**Attachments**
If applicable, add attachment files to reproduce the issue.

**Indigo/Bingo version**  
Version 1.7.2.0-gfd50d72ab-wasm32-wasm-clang-13.0.0

even1024:
Fixed in 1.7.4

PR #766, opened 2022-06-20
core: base64 support for binary formats, uninitialized data and #764 fix

even1024:


core/indigo-core/common/base_cpp/scanner.cpp
```diff
@@ -796,6 +806,27 @@ byte BufferScanner::readByte()
...
+        val = (val << 6) + T[c];
+        valb += 6;
+        if (valb >= 0)
+        {
+            out.push_back(char((val >> valb) & 0xFF));
+            valb -= 8;
+        }
+    }
+}
+
```
mkviatkovskii, line 829: There is no need in in this code, we already have cppcodec library in third_party directory, let's use it instead.

core/indigo-core/molecule/src/molecule_auto_loader.cpp
```diff
@@ -221,6 +240,22 @@ void MoleculeAutoLoader::_loadMolecule(BaseMolecule& mol, bool query)
...
+            loader.stereochemistry_options = stereochemistry_options;
+            if (query)
+                throw Error("CDX queries not supported yet");
+            loader.loadMolecule(mol.asMolecule());
+
+            properties.copy(loader.properties);
+
+            return;
+        }
+    }
```
mkviatkovskii, line 257: Why do we move it upper?

even1024: Binary formats should be handled first to avoid text format parsers of trying to parse binary.

core/indigo-core/common/base_cpp/scanner.cpp
even1024: fixed

mkviatkovskii (approved):

PR #772, opened 2022-07-01
Bingo Elastic search

AATDev21:
- Exact search was switched to use molecules hashes values for getting hits from Elastic.
- Changed substructure search query type.
- Added additional bad valence check into data importing step.
- Changed some Elastic search parameters.


MysterionRise:
we need to apply same changes in Java API as well

bingo/bingo-elastic/python/.pylintrc
```diff
@@ -4,6 +4,7 @@ disable=
     C0116, # missing-function-docstring
     C0115, # missing-class-docstring
     R0903, # too-few-public-methods
+    R0913, # too-many-arguments
```
mkviatkovskii, line 7: We should not add this exclusion. Probably it's better to refactor the code that raises this warning.

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -133,7 +134,9 @@ async def a_create_index(
     index_name: str, el_client: "AsyncElasticsearch"
 ) -> None:
     try:
-        await el_client.indices.create(index=index_name, body=index_body)
+        await el_client.indices.create(
+            index=index_name, body=index_body, ignore=400
```
mkviatkovskii, line 138: What does this `ignore=400` mean?

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -104,7 +104,8 @@ def get_client(
             "sim_fingerprint_len": {"type": "integer"},
             "sub_fingerprint": {"type": "keyword", "similarity": "boolean"},
             "sub_fingerprint_len": {"type": "integer"},
-            "cmf": {"type": "binary"},
+            "cmf": {"type": "keyword"},
+            "hash": {"type": "keyword", "similarity": "boolean"},
```
mkviatkovskii, line 108: Maybe it's easier to set hash type as integer?

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -151,13 +154,12 @@ def prepare(
 
 
 def response_to_records(
-    res, index_name, postprocess_actions
+    res, index_name, postprocess_actions, indigo_session=None, options=""
```
mkviatkovskii, line 157: Could you please add the type hints here?

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -315,14 +317,16 @@ def filter(
         self,
         similarity: Union[BaseMatch] = None,
         exact: IndigoRecord = None,
-        substructure: IndigoRecord = None,
-        limit=10,
+        substructure: IndigoObject = None,
+        indigo_session: Indigo = None,
+        limit: int = 5000,
```
mkviatkovskii, line 322: Why 5000?

bingo/bingo-elastic/python/bingo_elastic/model/record.py
```diff
@@ -29,37 +30,50 @@ def __set__(self, instance: IndigoRecord, value: Dict):
 
 class WithIndigoObject:
     def __set__(self, instance: IndigoRecord, value: IndigoObject) -> None:
+        value.aromatize()
```
mkviatkovskii, line 33: We can accidentially affect the state of the input object, probably it's better to clone it first.

bingo/bingo-elastic/python/bingo_elastic/queries.py
```diff
@@ -63,46 +63,41 @@ def compile(
...
     def clauses(self) -> List[Dict]:
-        return clauses(self._value.sub_fingerprint, "sub_fingerprint")
+        fp_list = self._value.fingerprint("sub").oneBitsList().split()
+        return clauses(fp_list, "sub_fingerprint")
 
     def compile(
         self, query: Dict, postprocess_actions: PostprocessType = None
     ) -> None:
         # This code same as ExactMatch.
         # ExactMatch will use search by hash in next releases
```
mkviatkovskii, line 93: This comment if obsolete, right?

bingo/bingo-elastic/python/bingo_elastic/queries.py
```diff
@@ -275,44 +270,36 @@ def __init__(self, target):
...
-            "source": "_score / doc['sub_fingerprint_len'].value"
-        }
-        query["min_score"] = 1
         assert postprocess_actions is not None
         postprocess_actions.append(getattr(self, "postprocess"))
 
 
 # Alias to default similarity match
-SimilarityMatch = TanimotoSimilarityMatch
+SimilarityMatch = EuclidSimilarityMatch
```
mkviatkovskii, line 302: Default should be Tanimoto

bingo/tests/dbc/BingoElastic.py
```diff
@@ -36,6 +38,8 @@ def import_data(self, data_path: str, database_type: str):
         index = 1
         for compound in indigo_iterator(self.indigo, data_path):
             try:
+                if compound.checkBadValence():
+                    raise IndigoException(compound.checkBadValence())
```
mkviatkovskii, line 42: We should leave it to user and his session settings. If he does not want to take account of the valence, then we also should not.

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -333,20 +337,19 @@ def filter(
...
 
 
 # pylint: disable=too-many-arguments
 def compile_query(
     similarity: BaseMatch = None,
     exact: IndigoRecord = None,
-    substructure: IndigoRecord = None,
-    limit: int = 10,
+    substructure: IndigoObject = None,
+    limit: int = 5000,
```
mkviatkovskii, line 349: Probably 100 as better as default value

bingo/bingo-elastic/python/bingo_elastic/elastic.py
MysterionRise: @mkviatkovskii it’s ignoring 400 http exceptions from Elasticsearch. Not sure why we would need to do this

bingo/bingo-elastic/python/bingo_elastic/elastic.py
MysterionRise: @AATDev21 @mkviatkovskii limit of 5000 would be substantial for Elasticsearch. We should stay with reasonable defaults. I’m okay with 20 but not 5000

bingo/bingo-elastic/python/bingo_elastic/elastic.py
MysterionRise: @AATDev21 @mkviatkovskii same as before. Default Elastic one is 10. We could override if needed. Getting 1 page of results of size 5000 is big load. Not sure for 100 too. Paging is much better way to deal with it

bingo/bingo-elastic/python/bingo_elastic/model/record.py
```diff
@@ -17,7 +17,8 @@ def check_error(instance: IndigoRecord, error: BaseException) -> None:
     if instance.error_handler:
         instance.error_handler(instance, error)
     else:
-        raise error
+        pass
```
MysterionRise, line 20: why removing raising of the error? If no error handler is provided - we shouldn’t “eat” it

bingo/bingo-elastic/python/bingo_elastic/model/record.py
```diff
@@ -29,37 +30,50 @@ def __set__(self, instance: IndigoRecord, value: Dict):
...
-                    .split(" ")
-                ]
-                setattr(instance, f"{f_print}_fingerprint", fp_)
-                setattr(instance, f"{f_print}_fingerprint_len", len(fp_))
+
+                fp_list = value.fingerprint(f_print).oneBitsList()
+                if fp_list:
+                    fp_ = [int(feature) for feature in fp_list.split(" ")]
+                    setattr(instance, f"{f_print}_fingerprint", fp_)
+                    setattr(instance, f"{f_print}_fingerprint_len", len(fp_))
```
MysterionRise, line 49: @AATDev21 @mkviatkovskii so if no fingerprint is created, we don’t set up some attributes. Is it happening often? Shouldn’t we throw some error that for this molecule we couldn’t generate fingerprint?

bingo/bingo-elastic/python/bingo_elastic/model/record.py
```diff
@@ -120,8 +135,8 @@ def as_dict(self) -> Dict:
         }
 
     def as_indigo_object(self, session: Indigo):
-        assert self.cmf
-        return session.deserialize(list(map(int, self.cmf.split(" "))))
+        assert self.cmf != ""
```
MysterionRise, line 138: @AATDev21 @mkviatkovskii @khyurri  I could think of this assert only if we aren’t sure that cml variable could be a string?

MysterionRise (changes requested):

bingo/bingo-elastic/python/bingo_elastic/elastic.py
AATDev21: In some test cases like substructure search for an empty query molecule, we expected to get 2026 hits, and a low limit value will lead this test to fail. I really shouldn't have changed limit's default value, so let's set it back to Elastic's default 10.

bingo/bingo-elastic/python/bingo_elastic/queries.py
AATDev21: Indeed, fixed it!

bingo/bingo-elastic/python/bingo_elastic/elastic.py
AATDev21: Explained giant limit below inside the similar comment and set it back to 10!

bingo/bingo-elastic/python/bingo_elastic/elastic.py
AATDev21: My bad, already added.

bingo/bingo-elastic/python/bingo_elastic/elastic.py
AATDev21: I added this parameter in order to get rid of long and ugly error logs "index already created". It's not neccessary now, so I removed it too, thanks!

bingo/tests/dbc/BingoElastic.py
AATDev21: I agree, but if, for example, a user wants to save all broken molecules, is it correctly to have them in the search results? And if not (our tests say - no), at what point it would be better to skip them? I decided to nip them in the bud, which is probably too harsh.

bingo/bingo-elastic/python/bingo_elastic/model/record.py
AATDev21: And in cases, when we expect it to be an actual empty string, catch deserialization error somehow else? I think the purpose of this assert originally was close to it.

bingo/bingo-elastic/python/bingo_elastic/model/record.py
AATDev21: Accidentally left here this "pass"( Thank you, brought everything back!

bingo/bingo-elastic/python/.pylintrc
AATDev21: Thanks, found a way to refactor and escape this warning!

bingo/bingo-elastic/python/bingo_elastic/model/record.py
AATDev21: The only reason not to have fingerprints is to be an empty molecule, right? I can't imagine any situation when we would need to have something like it in index, so probably we actually should. But what do you think we should do in this case with the empty molecule test?

bingo/bingo-elastic/python/bingo_elastic/model/record.py
khyurri: `assert self.cmf` and `assert self.cmf != ""` looks messy

I'd prefer something like that
```
def as_indigo_object(self, session: Indigo):
    if self.cmf:
        return session.deserialize(list(map(int, self.cmf.split(" "))))
    raise ValueError("Unexpected cmf value")
```

AATDev21: Fixed, thank you, Ruslan!

bingo/bingo-elastic/python/bingo_elastic/model/record.py
AATDev21: Fixed!

bingo/tests/dbc/BingoElastic.py
AATDev21: Removed this check from here and added as a filtration parameter to Elastic query.

MysterionRise (approved):
LGTM in general, apart from missing default similarity match? Should we have it?

mkviatkovskii (approved):

Issue #798, opened 2022-07-25
api: python: reorganize codebase, drop support of Python 2


mkviatkovskii:
**Is your feature request related to a problem? Please describe.**
Current implementation is a mess, we need to clear it, split into multiple packages and start adding type hints.

**Describe the solution you'd like**
1. Split `__init__.py` into multiple modules in `indigo` package
2. Move ctypes-related logic to standalone classes
3. Fix all flake8 warnings
4. Drop Python 2 support, set Python 3.6 as minimal supported version.
5. Start adding type hints.

PR #799, opened 2022-07-25
#798: api: python: reorganize codebase

mkviatkovskii:


api/python/indigo/bingo/bingo.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def close(self):
+        if self._id != -1:
+            IndigoLib.checkResult(
+                self._lib().bingoCloseDatabase(self._id), BingoException
+            )
+            self._id = -1
+
+    @staticmethod
+    def createDatabaseFile(indigo, path, databaseType, options=""):
```
AATDev21, line 51: Please add docstrings here like it was done for Inchi. I see most methods here are self-explanatory, but it would be way more convenient to have some reference with args/returns and a short description.

api/python/indigo/bingo/bingo_object.py
```diff
@@ -0,0 +1,131 @@
...
+        return self._db._lib()  # noqa
+
+    def close(self):
+        if self._id != -1:
+            IndigoLib.checkResult(
+                self._lib().bingoEndSearch(self._id), BingoException
+            )
+            self._id = -1
+
+    def next(self):
```
AATDev21, line 61: Same here (see bingo.py comment).

AATDev21 (changes requested):
Please add docstrings in a couple files and it will be fine.

api/python/indigo/bingo/bingo.py
mkviatkovskii: Good idea, but it's not the target of this request (it's just a refactoring).

Issue #520, opened 2021-11-11
core: replace own maps and sets implementation with std


mkviatkovskii:
**Motivation**
Currently Indigo uses own-written implementation of ordered associative containers and ordered sets using red-black trees. The implementation could be found in `core/indigo-core/common/base_cpp/red_black.h`.
In many cases order of elements is not important and `int` is a key type, so it could be easily replaced with hash maps and also lead to performance improvement. Same for unordered sets.

**ToDo**
1. Find all places in code that use `RedBlackMap<K, V>`, check if K is trivially hashable and order of elements is required, and replace it with `std::map<K, V>` or `std::unordered_map<K, V>` depending on the situation.
2. Find all places in code that use `RedBlackSet<K>`, check if K is trivially hashable and order of elements is required, and replace it with `std::set<K>` or `std::unordered_set<K>` depending on the situation.
3. Check if `core/indigo-core/common/base_cpp/red_black.h` and eliminate it if it's not required anymore.

PR #818, opened 2022-08-04
Issue #520: replace RedBlackMap for MoleculeAlleneStereo

AStepaniuk:
Issue #520: replace RedBlackMap and RedBlackSet implementation with standard containers
Partial implementation: fix for MoleculeAlleneStereo

### DIFF
--- a/core/indigo-core/molecule/molecule_allene_stereo.h
+++ b/core/indigo-core/molecule/molecule_allene_stereo.h
@@ -24,6 +24,9 @@
 #pragma warning(disable : 4251)
 #endif
 
+#include <functional>
+#include <map>
+
 #include "base_cpp/exception.h"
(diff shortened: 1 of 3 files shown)


core/indigo-core/molecule/molecule_allene_stereo.h
```diff
@@ -86,7 +86,7 @@ namespace indigo
 
         bool _isAlleneCenter(BaseMolecule& mol, int idx, _Atom& atom, int* sensible_bonds_out);
 
-        RedBlackMap<int, _Atom> _centers;
+        std::map<int, _Atom> _centers;
```
mkviatkovskii, line 89: That's what we discussed. We can't use standard containers as members of classes that are used inside Indigo memory-based containers. It leads to memory corruption. That's why the tests failed.

Issue #779, opened 2022-07-18
Get rid of all shallow copy operations at array.h


even1024:
Use incapsulated vector operations instead of reallocs and memcpy. vector<>::reserve, vector<>::resize and std::copy should be used instead.

PR #821, opened 2022-08-05
Issue #779: remove NonCopyable and shallow copy operations in Array<>

even1024:


core/indigo-core/molecule/src/cml_loader.cpp
```diff
@@ -18,12 +18,11 @@
 
 #include "molecule/cml_loader.h"
 
-#include <tinyxml2.h>
-
 #include "base_cpp/scanner.h"
 #include "molecule/elements.h"
 #include "molecule/molecule.h"
 #include "molecule/molecule_scaffold_detection.h"
+#include <tinyxml2.h>
```
mkviatkovskii, line 25: That's incorrect change, we have three levels of headers, splitted by empty line:

1. standard library
2. third-party libraries
3. our code

Issue #815, opened 2022-08-04
Replace ObjArray with std::vector


AStepaniuk:
ObjArray is an object-wise wrapper around Array.
Though ObjArray is still not safe to use with C++ objects because underlying Array does not properly copies/moves objects on buffer re-allocation.

A simple solution here: replace ObjArray with std::vector, which has pretty similar interface and correctly deals with contained objects.
Challenges:
1. The contained type needs to have move constructor, marked as noexcept
2. ObjArray can `qsort()` items using its own method. This method needs to be extracted as a standalone function, implemented as an algorithm

PR #824, opened 2022-08-10
#815 Replace ObjArray with std::vector.

AStepaniuk:
#815 Replace ObjArray with std::vector.
ObjArray is removed from PropertiesMap class. std classes are used instead.


AliaksandrDziarkach (dismissed):
LGTM

PR #864, opened 2022-10-07
Update to jdk 17 and various libs

MysterionRise:


MysterionRise:
Not relevant due to pharma industry still heavy relying on Java 8

Issue #874, opened 2022-10-13
Support generic and polymer s-groups for extended SMILES


even1024:
Polymer and generic s-groups information should be parsed in extended SMILES parser.
Also export to extended SMILES of the structures with those s-groups should keep  the information.
There are basic examples:

CCCC |Sg:gen:0,1,2:|
CCCC |Sg:n:0,1,2:3-6:eu|

PR #875, opened 2022-10-13
S-Groups support for extended SMILES (#874)

even1024:


api/tests/integration/tests/formats/smiles.py
```diff
@@ -86,3 +86,11 @@
         print("  -> " + sm3)
     except IndigoException as e:
         print(getIndigoExceptionText(e))
+
+print("*** S-Groups ***")
+mols_smiles = [
+    "CCCC |Sg:gen:0,1,2:|",
+    "CCCC |Sg:n:0,1,2:3-6:eu|",
+]
+for sm in mols_smiles:
```
AlexanderSavelyev, line 95: sgroups should not be saved in daylight smiles

core/indigo-core/molecule/src/smiles_saver.cpp
```diff
@@ -1992,6 +2042,9 @@ void SmilesSaver::_checkSRU()
 
     int i, j, k;
 
+    if (write_extra_info) // SRU will be handled better inside the extended block
+        return;
```
AlexanderSavelyev, line 2046: please disable code below. what does "write_extra_info" mean. Is it extended block?

AlexanderSavelyev (approved):

Issue #868, opened 2022-10-10
Add parameter for dearomatize on load


evgeniiFrolov1:
**Is your feature request related to a problem? Please describe.**
In the ketcher we would like to add the property `dearomatize`. The main goal of this feature is to add a parameter to `layout` and to `convert` API to dearomatize struct on load automatically

**Describe the solution you'd like**
To add a parameter to `layout` and to `convert` API to dearomatize struct on load automatically

AlexanderSavelyev:
make indigo option "dearomatize-on-load"

even1024:
fixed in 1.8.0.rc.2

PR #879, opened 2022-10-17
Dearomatize on load #868

even1024:


api/c/indigo/src/indigo_molecule.cpp
```diff
@@ -508,12 +510,14 @@ CEXPORT int indigoLoadQueryMolecule(int source)
...
         loader.treat_x_as_pseudoatom = self.treat_x_as_pseudoatom;
+        loader.dearomatize_on_load = self.dearomatize_on_load;
+        loader.arom_options = self.arom_options;
 
         std::unique_ptr<IndigoQueryMolecule> molptr = std::make_unique<IndigoQueryMolecule>();
 
         QueryMolecule& qmol = molptr->qmol;
 
-        loader.loadQueryMolecule(qmol);
+        loader.loadMolecule(qmol);
```
mkviatkovskii, line 520: ?

api/c/indigo/src/indigo_reaction.cpp
```diff
@@ -383,9 +385,11 @@ CEXPORT int indigoLoadQueryReaction(int source)
 
         loader.stereochemistry_options = self.stereochemistry_options;
         loader.treat_x_as_pseudoatom = self.treat_x_as_pseudoatom;
+        loader.dearomatize_on_load = self.dearomatize_on_load;
+        loader.arom_options = self.arom_options;
 
         std::unique_ptr<IndigoQueryReaction> rxnptr = std::make_unique<IndigoQueryReaction>();
-        loader.loadQueryReaction(rxnptr->rxn);
+        loader.loadReaction(rxnptr->rxn);
```
mkviatkovskii, line 392: ?

api/c/indigo/src/indigo_molecule.cpp
even1024: loadMolecule now can load both: queries and molecules. Actually we don't need separate methods to do that.

api/c/indigo/src/indigo_reaction.cpp
even1024: We don't need separate loadQueryReaction. loadReaction can handle both types.

core/indigo-core/molecule/molecule_auto_loader.h
```diff
@@ -38,6 +39,13 @@ namespace indigo
     class QueryMolecule;
     class BaseMolecule;
 
+    enum AutoAromatizeMode
+    {
+        NO_AUTO_AROMATIZE = 0,
+        AUTO_AROMATIZE,
+        AUTO_DEAROMATIZE
+    };
```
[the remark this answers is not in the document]
mkviatkovskii: Also. if we already define AUTO_AROMATIZE, let's add this option as well. Let's add option "auto-aromatizon-mode" with options "disabled", "aromatize" and "dearomatize", that would be very useful.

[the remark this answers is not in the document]
even1024: enum removed

PR #882, opened 2022-10-25
Feature/861 cdxml enhanced stereochemistry

even1024:
Enhanced stereochemistry support for cdxml format. (Import/Export)


core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
```diff
@@ -141,6 +143,68 @@ void MoleculeCdxmlLoader::_parseCollections(BaseMolecule& mol)
...
+void MoleculeCdxmlLoader::_processEnhancedStereo(BaseMolecule& mol)
+{
+    std::vector<int> ignore_cistrans(mol.edgeCount());
+    std::vector<int> sensible_bond_directions(mol.edgeCount());
+    for (int i = 0; i < mol.edgeCount(); i++)
+        if (mol.getBondDirection(i) == BOND_EITHER)
+        {
+            if (MoleculeCisTrans::isGeomStereoBond(mol, i, 0, true))
+            {
+                ignore_cistrans[i] = 1;
```
AlexanderSavelyev, line 155: what is 1. is it constant

core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
```diff
@@ -141,6 +143,68 @@ void MoleculeCdxmlLoader::_parseCollections(BaseMolecule& mol)
...
+                sensible_bond_directions[i] = 1;
```
AlexanderSavelyev, line 156: what is 1

AlexanderSavelyev (commented):
don't see cdxml_to_mol.py changes

core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
even1024: 1 means "true" here. we use std::vector<int> because std::vector<bool> uses optimized inner data representation, but indigo methods like BaseMolecule::buildCisTrans require plain int data array where 1 means true and 0 means false. BTW according to C++11 specification true converts to 1 during casting to int. So changed 1 to the true.

core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
even1024: changed 1 to the true

AlexanderSavelyev (approved):

PR #890, opened 2022-11-07
Feature/#862 cdxml abbreviations

even1024:
Superaroms support


api/tests/integration/ref/formats/mol_to_cdxml.py.out
```diff
@@ -19,49 +20,49 @@ enhanced_stereo1.mol
...
-            <n id="5" Element="6" p="110.021996 168.599991"/>
-            <n id="6" Element="6" p="150.029999 191.699997"/>
-            <n id="7" Element="6" p="150.029999 237.899994"/>
-            <n id="8" Element="6" p="110.021996 261.000000"/>
-            <n id="9" Element="6" p="70.011002 237.899994"/>
-            <n id="10" Element="6" p="70.011002 191.699997"/>
-            <n id="11" Element="7" NumHydrogens="1" p="30.000000 122.399994">
+            <n id="5" p="70.011002 53.099998"/>
+            <n id="6" p="70.011002 99.299995" Geometry="Tetrahedral" BondOrdering="5 7 14 4" EnhancedStereoType="Absolute"/>
+            <n id="7" p="110.021996 122.399994"/>
```
AlexanderSavelyev, line 31: I can see that Element was removed. Just double check - is it expected

core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
```diff
@@ -111,8 +133,14 @@ void MoleculeCdxmlLoader::_parseCollections(BaseMolecule& mol)
...
-        case kCDXNodeType_ExternalConnectionPoint:
-            break;
+        case kCDXNodeType_ExternalConnectionPoint: {
+
+            auto& fn = nodes[_fragment_nodes.back()];
+            if (fn.connections.size() == 0)
+                fn.ext_connections.push_back(node.id);
+        }
+        break;
+        case kCDXNodeType_Nickname:
```
AlexanderSavelyev, line 143: is it ok that case without break?

core/indigo-core/reaction/src/reaction_cdxml_saver.cpp
```diff
@@ -387,8 +387,22 @@ void ReactionCdxmlSaver::_addStep(BaseReaction& rxn, MoleculeCdxmlSaver& molsave
...
+    // generate ids for meta objects. 1 node and 1 extra object. text or graphics
+    for (auto i = 0; i < rxn.meta().metaData().size(); ++i)
+    {
+        int r_id = i + rxn.end();
+        id++;
+        meta_ids.insert(std::make_pair(r_id, id));
+        nodes_ids.expand(r_id + 1);
+        nodes_ids[r_id].clear_resize(1);
+        nodes_ids[r_id].zerofill();
+        id += 2;
```
AlexanderSavelyev, line 402: why id++ above and id+=2?

core/indigo-core/molecule/src/molecule_cdxml_loader.cpp
even1024: break statement can be placed inside brackets of 'case' or outside. This is a question of code style. Both variants work with the same manner.

core/indigo-core/reaction/src/reaction_cdxml_saver.cpp
even1024: There is a comment above:     // generate ids for meta objects. 1 node and 1 extra object. text or graphics
Every metaobject requires two ids.  First is a meta object id itself and the second one for nested text or graphics element so we can easily guess the nested block id from meta object id.

api/tests/integration/ref/formats/mol_to_cdxml.py.out
even1024: fixed

AlexanderSavelyev (approved):

PR #904, opened 2022-11-24
Bugfix/896 cdxml export multistep

even1024:
Fixes multistep reactions CDXML import/export issues.


utils/indigo-depict/main.c
```diff
@@ -955,7 +955,7 @@ int main(int argc, char* argv[])
         _prepare(obj, p.aromatization);
         if (p.action == ACTION_LAYOUT)
         {
-            indigoLayout(obj);
+            //indigoLayout(obj);
```
mkviatkovskii, line 958: This file probably should not be changed.

even1024: fixed

mkviatkovskii (approved):

Issue #521, opened 2021-11-11
core: replace MultiMap with standard library container


mkviatkovskii:
**Motivation**
Currently Indigo uses own-written implementation of MultiMap instead of standard one. It does not contain any unique API that could not be replaced with standard container. Also it uses red-black trees even if we don't need the order of the keys, so we could also achieve performance improvement here.

**ToDo**
1. Find all places in code that use `MultiMap<K, V>`, check if K is trivially hashable and order of elements is required, and replace it with `std::multimap<K, V>` or `std::unordered_multimap<K, V>` depending on the situation.

PR #917, opened 2022-11-30
Issue #521: core: replace MultiMap in MoleculeRGroupsComposition class

loimu:
Partial implementation: replace MultiMap with standard library container


MysterionRise (approved):
LGTM, changes in out file are because of different ordering?

loimu:
> LGTM, changes in out file are because of different ordering?

Correct, nothing is changed besides the order of fragments

PR #921, opened 2022-12-05
Partial implementation: replace MultiMap in MoleculeRGroupsComposition class - Part 2, remove MultiMap completely

loimu:
Partial implementation: replace MultiMap with standard library container


MysterionRise:
@loimu could you update it?

PR #930, opened 2022-12-14
Feature/885 cdx support

even1024:
CDX format support


mkviatkovskii (changes requested):
Commenting failing tests is obviously not a good way to fix failing tests.

PR #947, opened 2022-12-22
Incorrect display of utf8 text in standalone #889

IvanMinin:


CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ option(WITH_STATIC "Build Indigo static library as well as shared" OFF)
 
 # Indigo API options
 option(BUILD_INDIGO "Build indigo shared library" ON)
+option(USE_FONT_MANAGER "Build indigo render with font_face_manager" OFF)
```
mkviatkovskii, line 18: We already have BUILD_STANDALONE for that. I would suggest to use font manager in case of standalone build.

third_party/CMakeLists.txt
```diff
@@ -9,11 +9,10 @@ if (BUILD_INDIGO OR BUILD_INDIGO_UTILS)
     if (BUILD_STANDALONE)
         add_subdirectory(libpng)
         add_subdirectory(pixman)
-        if (EMSCRIPTEN)
-            add_subdirectory(cairo-wasm)
-        else()
-            add_subdirectory(cairo)
+        if (UNIX AND NOT APPLE)
+            add_subdirectory(freetype)
```
mkviatkovskii, line 13: We also need to add possibility to use system freetype on Linux (see cairo or libpng or libz as an example).

core/render2d/src/render_font_face_manager.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    RenderFontFaceManager::~RenderFontFaceManager()
+    {
+        // TODO: fix faces realease
+        // TODO: fix function signature mismatch in _cairo_user_data_array_fini
+        // cairo_font_face_destroy (_cairo_face_regular);
+        // cairo_font_face_destroy (_cairo_face_bold);
+        // cairo_font_face_destroy (_cairo_face_italic);
+        // cairo_font_face_destroy (_cairo_face_bold_italic);
```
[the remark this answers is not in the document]
IvanMinin: done

CMakeLists.txt
IvanMinin: Right now, we want to use font manager only in case of WASM.
It will be added later to standalone.

third_party/CMakeLists.txt
IvanMinin: Fixed.
In case of NOT EMSCRIPTEN system freetype and fontconfig libs will be used.

mkviatkovskii (approved):

Issue #950, opened 2022-12-23
CDX import: Aromatized structures are not recognized when Pasting from Clipboard


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Paste into Clipboard
```
VmpDRDAxMDAEAwIBAAAAAAAAAAAAAACAAAAAAAMAFQAAAENoZW1EcmF3IEpTIDIu
MC4wLjkEAhAAeIf0AFtqMAKHeFMB+eq2AgEJCAAAAAAAAAAAAAIJCAAAAGwDAABb
Bw0IAQABCAcBAAE6BAEAATsEAQAARQQBAAE8BAEAAEoEAQAADAYBAAEPBgEAAQ0G
AQAAQgQBAABDBAEAAEQEAQAADggCAJoCCggIABgAYADIAAMACwgIABgAAADIAAMA
CQgEAACAAgAICAQAmZkBAAcIBACZmQAABggEAAAAAgAFCAQAZmYOAAQIAgC0AAMI
BAAAAHgAIwgBAAUMCAEAACgIAQABKQgBAAEqCAEAATIIAQAAKwgBACgsCAEACi0I
AQABLggBAAACCBAAAAAkAAAAJAAAACQAAAAkAAEDAgAAAAIDAgABAAADMgAIAP//
/////wAAAAAAAP//AAAAAP////8AAAAA//8AAAAA/////wAAAAD/////AAD//wAB
DwAAAAEAGADp/QUAQXJpYWwACHgAAAMAAAEgASAAAAAAC2YIoP+E/4gL4wkYA2cF
JwP8AAIAAAEgASAAAAAAC2YIoAABAAAAZAAAAAEAAQEBAAAAAScPAAEAAQAAAAAA
AAAAAAAAAAACABkBkAAAAAAAQAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAtAsC
AAAAtQsUAAAAQ2hlbWljYWwgRm9ybXVsYTogtgsOAAAARXhhY3QgTWFzczogtwsU
AAAATW9sZWN1bGFyIFdlaWdodDoguAsHAAAAbS96OiC5CxYAAABFbGVtZW50YWwg
QW5hbHlzaXM6ILoLEQAAAEJvaWxpbmcgUG9pbnQ6ILsLEQAAAE1lbHRpbmcgUG9p
[...]

PR #953, opened 2022-12-26
CDX import: Aromatized structures are not recognized when Pasting from Clipboard #950

even1024:


core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
```diff
@@ -1520,6 +1520,11 @@ void MoleculeCdxmlSaver::endDocument()
 {
     if (_is_binary)
     {
+        std::ifstream t("C:\\cdx\\benz.xml");
```
mkviatkovskii, line 1523: That is definitely something that should not exist in this pull request.

even1024: removed

Issue #928, opened 2022-12-13
Error opening MOL and RXN files with RBC/SUB/UNC queries


Zhirnoff:
Ketcher fails to open MOL V3000 and RXN v3000 files with RBC/SUB/UNC features

**Steps to Reproduce**
1. Open Ketcher
2. Create or open attached [MOL v3000](https://github.com/epam/Indigo/files/10218083/queryv3000.zip) and [RXN v3000](https://github.com/epam/Indigo/files/10320309/marvinjs_untitled_file.2.zip) with RBC/SUB/UNS queries

3. Look at error message

**Expected behavior**
Attached file added to canvas without error messages

**Actual behavior**
Remote Error message: Convert error! struct data not recognized as molecule, query, reaction or reaction query
Standalone Error message: Convert error! Given string could not be loaded as (query or plain) molecule or reaction, see the error messages: 'molfile loader: unsaturation flag is allowed only for queries', 'RXN loader: bad header ', 'RXN loader: bad header ', 'molfile loader: invalid RBCNT value: -2'

**Attachments**

![2022-12-13_16h12_30](https://user-images.githubusercontent.com/62464241/207328001-df8f0729-05df-4bd4-856f-ac955144066c.png)
![2022-12-13_16h12_50](https://user-images.githubusercontent.com/62464241/207328019-04032cc4-7fac-4853-ad75-cf29f2b08718.png)

**Indigo/Bingo version**  
Version 1.8.3.0-gded13d6dc-x86_64-linux-gnu-11.2.1
Ketcher related issue: https://github.com/epam/ketcher/issues/1918

even1024:
fixed in https://github.com/epam/Indigo/pull/969

PR #969, opened 2023-01-04
Error opening MOL and RXN files with RBC/SUB/UNC queries #928

even1024:


AlexanderSavelyev (commented):
please add tests for substructure matcher for subst=-2

Issue #960, opened 2023-01-03
CDX Import, CDXML Import: parsing error when superatom starts with 'R' symbol


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Paste from Clipboard or open [File](https://github.com/epam/Indigo/files/10336207/marvinjs_untitled_file.80.zip)

```
[...]

PR #975, opened 2023-01-09
CDX Import, CDXML Import: parsing error when superatom starts with 'R' symbol #960

even1024:
https://github.com/epam/Indigo/issues/960


core/indigo-core/molecule/molecule_cdxml_loader.h
```diff
@@ -737,20 +739,30 @@ namespace indigo
             return _size ? getBinaryValue() : std::string(xml().Value());
         }
 
-        std::string getBinaryValue()
+        std::string getBinaryName()
```
IvanMinin, line 742: We can make this function const.

even1024: fixed

Issue #979, opened 2023-01-10
CDXML import fails to load rectangle primitives


even1024:
Example attached.

[test.zip](https://github.com/epam/Indigo/files/10381003/test.zip)

PR #980, opened 2023-01-10
CDXML import fails to load rectangle primitives #979

even1024:


utils/indigo-depict/main.c
```diff
@@ -907,7 +907,7 @@ int main(int argc, char* argv[])
         _prepare(obj, p.aromatization);
         if (p.action == ACTION_LAYOUT)
         {
-            //indigoLayout(obj);
```
AlexanderSavelyev, line 910: why this layout is uncomment

IvanMinin:
Only a test is added in this pull request.
 The actual fix was added in the another pull request.

IvanMinin (approved):
Only a test is added in this pull request.
The actual fix was added in the another pull request.

Issue #963, opened 2023-01-03
CDXML: File containing Functional Groups or Salts and Solvents cannot be opened and causes a convert error


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Create structure with Functional Groups or Salts and Solvents ([ketcher (35).zip](https://github.com/epam/Indigo/files/10338106/ketcher.35.zip))
3. Save it as CDXML
4. Open from file or Paste from Clipboard

**Expected behavior**
File containing Functional Groups or Salts and Solvents can be opened witout convert error.

**Actual behavior**
File containing Functional Groups or Salts and Solvents cannot be opened and causes a convert error

**Attachments**

https://user-images.githubusercontent.com/62464241/210397549-28d4675b-d386-43cf-885b-6fc5734c9c92.mp4

**Indigo/Bingo version**  
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.9.0-rc.2.0-g40c4e3793-x86_64-linux-gnu-11.2.1

**Additional context**

PR #984, opened 2023-01-11
CDXML: File containing Functional Groups or Salts and Solvents cannot be opened and causes a convert error #963

even1024:
https://github.com/epam/Indigo/issues/963


core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
```diff
@@ -885,7 +885,7 @@ void MoleculeCdxmlSaver::addFragmentNodes(BaseMolecule& mol, tinyxml2::XMLElemen
     {
         std::vector<std::pair<int, int>> ext_connections;
         std::vector<int> connection_order, bond_ordering;
-        std::vector<int> int_connections;
+        std::set<int> int_connections;
```
IvanMinin, line 888: Why is std::set  used instead of std::unordered_set?

core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
```diff
@@ -917,7 +917,10 @@ void MoleculeCdxmlSaver::addFragmentNodes(BaseMolecule& mol, tinyxml2::XMLElemen
                 }
 
                 if (_bonds_included.find(nei_edge_idx) != _bonds_included.end())
-                    int_connections.push_back(nei_edge_idx);
+                {
+                    if (int_connections.find(nei_edge_idx) == int_connections.end())
```
IvanMinin, line 921: Do we need this check or is it just for readability?

IvanMinin (approved):

Issue #990, opened 2023-01-13
CDXML import: nodes with radicals are not getting parsed


even1024:
For CDXML nodes like:

<n
 id="7"
 p="850.50 404.80"
 Z="7"
 NumHydrogens="0"
 Radical="Doublet"
 NeedsClean="yes"
 AS="N"
>

Radicals should be converted to the corresponding indexes as specified:

enum CDXRadical
{ // Same as MDL codes
    kCDXRadical_None = 0,
    kCDXRadical_Singlet = 1, // diradical singlet  (two dots)
    kCDXRadical_Doublet = 2, // monoradical		  (one dot)
    kCDXRadical_Triplet = 3  // diradical triplet  (two dots)
};

[benzene_radical.cdxml.txt](https://github.com/epam/Indigo/files/10412611/benzene_radical.cdxml.txt)

PR #991, opened 2023-01-13
CDXML import: nodes with radicals are not getting parsed #990

even1024:
https://github.com/epam/Indigo/issues/990


IvanMinin:
I suggest adding test case for other values of "Atom_Radical Property" (None, Singlet, Triplet).

IvanMinin (approved):
I suggest adding test case for other values of "Atom_Radical Property" (None, Singlet, Triplet).

Issue #941, opened 2022-12-19
cdx export


even1024:
Implement export to CDX format, add corresponding functions to Indigo API including WASM and remote Indigo service.

even1024:
implemented in 1.10-rc.1 
https://github.com/epam/Indigo/pull/1008

PR #1008, opened 2023-01-26
cdx export #941

even1024:


utils/indigo-depict/main.c
```diff
@@ -973,10 +977,14 @@ int main(int argc, char* argv[])
                 indigoSaveRxnfileToFile(obj, p.outfile);
             else if (p.out_ext == OEXT_CDXML || p.out_ext == OEXT_CDXMLR)
                 indigoSaveCdxmlToFile(obj, p.outfile);
+            else if (p.out_ext == OEXT_CDX || p.out_ext == OEXT_CDR)
+            {
+                indigoSaveCdxToFile(obj, p.outfile);
+                printf("cdx: %s\n", indigoCdx(obj));
```
AlexanderSavelyev, line 983: remove debug print

api/c/indigo/indigo.h
```diff
@@ -203,6 +203,7 @@ CEXPORT int indigoSaveJson(int item, int output);
 CEXPORT int indigoSaveCml(int object, int output);
 CEXPORT int indigoSaveCmlToFile(int object, const char* filename);
 CEXPORT const char* indigoCml(int object);
+CEXPORT const char* indigoCdx(int object);
```
AlexanderSavelyev, line 206: is CDX is a string?

AlexanderSavelyev: if it is base64 - it needs to be renamed to indigoBas64Cdx

api/c/indigo/src/indigo_object.cpp
```diff
@@ -16,14 +16,14 @@
...
-#include "indigo_internal.h"
-
-#include <map>
-
 #include "base_cpp/output.h"
 #include "base_cpp/properties_map.h"
+#include "indigo_internal.h"
 #include "reaction/reaction.h"
 
+#include <../cppcodec/cppcodec/base64_default_rfc4648.hpp>
```
AlexanderSavelyev, line 24: do we have a include path for cppcodec?

api/python/indigo/indigo/indigo_object.py
```diff
@@ -305,6 +319,15 @@ def cdxml(self):
 
         return IndigoLib.checkResultString(self._lib().indigoCdxml(self.id))
 
+    def cdx(self):
```
AlexanderSavelyev, line 322: need to rename because it is not cdx

api/python/indigo/indigo/indigo_object.py
```diff
@@ -305,6 +319,15 @@ def cdxml(self):
...
+        """Molecule method returns the structure as a string in CDX base64 encoded format
+
+        Returns:
+            str: CDXML string
```
AlexanderSavelyev, line 326: fix comment

api/python/indigo/indigo/indigo_object.py
AlexanderSavelyev: is it base64Cdx?

utils/indigo-service/backend/service/v2/indigo_api.py
```diff
@@ -373,6 +373,8 @@ def save_moldata(md, output_format=None, options={}, indigo=None):
         return md.struct.cml()
     elif output_format == "chemical/x-cdxml":
         return md.struct.cdxml()
+    elif output_format == "chemical/x-cdx":
```
AlexanderSavelyev, line 376: do we have format for base64-cdx?

api/c/indigo/src/indigo_internal.h
```diff
@@ -200,6 +200,8 @@ class DLLEXPORT IndigoObject
     virtual MonomersProperties& getMonomersProperties();
     virtual void copyProperties(PropertiesMap&);
 
+    void toBase64String(Array<char>& str);
```
AlexanderSavelyev, line 203: do we use this function anywhere? if yes, need to change to return result array instead of converting input to base64

AlexanderSavelyev (changes requested):
need to add tests

Issue #957, opened 2022-12-28
Support of Korean, Chinese and Japanese characters in Standalone.


IvanMinin:
Add support of Korean, Chinese and Japanese characters to wasm indigo-ketcher.

PR #1027, opened 2023-02-07
#957 Support of Korean, Chinese and Japanese characters in Standalone

IvanMinin:


IvanMinin:
Updated NotoSans fonts and tests.

Added CJK regular and bold fonts and tests.

Added lazy loading of fonts.

Removed static variables 'key'.

api/wasm/indigo-ketcher/CMakeLists.txt
```diff
@@ -53,6 +53,23 @@ add_custom_target(${PROJECT_NAME}-test
         DEPENDS ${PROJECT_NAME}
         )
 
+if(RENDER_ENABLE_CJK)
+        add_custom_target(${PROJECT_NAME}-test-cjk
+        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/test_cjk.js ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/
```
AlexanderSavelyev, line 58: why do we copy tests inside cmake build?

IvanMinin: This file is copied to run tests inside "build" folder.
The same way as is it done for no CJK tests.

AlexanderSavelyev: can we remote hardcoded tests name from the cmake?

IvanMinin: yes, done

AlexanderSavelyev (approved):

Issue #965, opened 2023-01-04
MDL Molfile v3000: when opening files containing 'Salts an Solvents', names are truncated and abbreviation is expanded


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Put any "Salts and Solvents" whose names contain spaces on canvas
3. Save as MDL Mol v3000
4. Open or Paste from Clipboard
Note: In MDL Molfile v3000 the name of the superatoms must be quoted if there are spaces.
Abbreviations are expanded because the name changes and they become unidentified.
![2023-01-04_18h06_31](https://user-images.githubusercontent.com/62464241/210585606-c0a01f89-26f2-4a94-b701-495c85262642.png)
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/62464241/210585768-ea87f95d-e9a1-40b0-bc85-bdd5920380b7.png)

**Expected behavior**
When opening files containing 'Salts an Solvents', the names are not truncated at the first space in the name and the abbreviation is parsed as it was saved

**Actual behavior**
When opening files containing 'Salts an Solvents', the names are truncated at the first space in the name and the abbreviation is parsed expanded

**Attachments**

https://user-images.githubusercontent.com/62464241/210584574-ef2c765a-f4bf-4699-9d5f-b3a6f01a0464.mp4

**Indigo/Bingo version**  
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.9.0-rc.2.0-g40c4e3793-x86_64-linux-gnu-11.2.1

**Additional context**
Related Ketcher issue: https://github.com/epam/ketcher/issues/2143

Zhirnoff:
The second half of the Salts and Solvents name is lost and quotes appear in the first part of the name.

https://user-images.githubusercontent.com/62464241/236281768-1890f5fc-16e8-4346-b4ed-cdb974016273.mp4

PR #1046, opened 2023-02-21
#965 MDL Molfile v3000: when opening files containing 'Salts an Solvents', names are truncated and abbreviation is expanded

even1024:


IvanMinin (approved):
Approved

Issue #1032, opened 2023-02-13
Combine molecules that are related to a single s-group into one in .Ket format


KonstantinEpam23:
**Background**
At the moment Indigo is unable to serialize a structure that contains s-groups that span several molecules into .Ket format correctly. It produces separate s-groups for each molecule.

For example, given this structure in .mol format (one s-group includes two separate atoms):
```
  -INDIGO-02092309082D

  0  0  0  0  0  0  0  0  0  0  0 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 2 0 1 0 0
M  V30 BEGIN ATOM
M  V30 1 O 11.75 -4.2 0.0 0
M  V30 2 N 13.1 -4.475 0.0 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 END BOND
M  V30 BEGIN SGROUP
M  V30 1 SUP 1 ATOMS=(2 1 2) BRKXYZ=(9 0.000000 0.000000 0.000000 0.000000 0.-
M  V30 000000 0.000000 0.000000 0.000000 0.000000) BRKXYZ=(9 0.000000 0.00000-
M  V30 0 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000) LABE-
M  V30 L=test
M  V30 END SGROUP
M  V30 END CTAB
M  END
```

Indigo currently produces the following output, inserting an s-group block into each molecule block, which is incorrect:
```
{
  "root": {
    "nodes": [ { "$ref": "mol0" }, { "$ref": "mol1" } ]
  },
  "mol0": {
    "type": "molecule",
    "atoms": [ { "label": "O", "location": [ 11.75, -4.199999809265137, 0.0 ] } ],
    "bonds": [],
    "sgroups": [ { "type": "SUP", "atoms": [ 0 ], "name": "sgroup test" } ]
  },
  "mol1": {
    "type": "molecule",
    "atoms": [ { "label": "N", "location": [ 13.100000381469727, -4.474999904632568, 0.0 ] } ],
    "bonds": [],
    "sgroups": [ { "type": "SUP", "atoms": [ 0 ], "name": "sgroup test" } ]
  }
}
```

**Describe the solution you'd like**
When an s-group includes multiple molecules - these molecules will be combined into a single one.
For the example .mol file above the resulting `.ket` serialization will look like this:
```
{
  "root": {
    "nodes": [ { "$ref": "mol0" } ]
  },
  "mol0": {
[...]

PR #1047, opened 2023-02-21
Backmerge: #1032 Combine molecules that are related to a single s-group into one in .Ket format

even1024:


IvanMinin (approved):
Approved

Issue #310, opened 2021-01-21
Support stereo CIP calculation in Ket format


AlexanderSavelyev:
**Acceptance criteria:**
Ket format should support CIP labels for atoms
Specification is updated
CIP labels should be moved from Sgroup -> Atom properties

PR #1054, opened 2023-03-02
#310 Support stereo CIP calculation in Ket format

even1024:


core/indigo-core/molecule/src/molfile_saver.cpp
```diff
@@ -149,6 +149,7 @@ void MolfileSaver::_saveMolecule(BaseMolecule& mol, bool query)
 
     MoleculeCIPCalculator mcc;
     mcc.updateCIPStereoDescriptors(mol, add_stereo_desc);
+    mcc.addCIPSgroups(mol);
```
AlexanderSavelyev, line 152: s-group should be added only if option is on

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/molfile_saver.cpp
even1024: updateCIPStereoDescriptors adds corresponding properties to bonds and atoms if add_stereo_desc = true.
addCIPSgroups has not effect if no properties has been added.

core/indigo-core/molecule/src/molfile_saver.cpp
```diff
@@ -149,6 +149,7 @@ void MolfileSaver::_saveMolecule(BaseMolecule& mol, bool query)
 
     MoleculeCIPCalculator mcc;
     mcc.updateCIPStereoDescriptors(mol, add_stereo_desc);
```
AlexanderSavelyev, line 151: change the function to be more obvious
if (add_stereo_desc) {
MoleculeCIPCalculator mcc;
mcc.addCIP....()
mcc.addCIPSgroups(mol);
}

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/molfile_saver.cpp
even1024: fixed

AlexanderSavelyev (approved):

Issue #1064, opened 2023-03-14
Keep implicit hydrogens information in KET-format


even1024:
**Steps to Reproduce**
1. Open attached MOL [test.mol.txt](https://github.com/epam/Indigo/files/10965988/test.mol.txt)
2. Save as KET
3. Open saved KET
4. Save it as MOL

**Expected behavior**
MRV_IMPLICIT_H should be saved.

**Actual behavior**
MRV_IMPLICIT_H is missing.

even1024:
[[](url)](https://github.com/epam/ketcher/issues/1865)

even1024:
#https://github.com/epam/ketcher/issues/1865

even1024:
#https://github.com/epam/ketcher/issues/1865

PR #1065, opened 2023-03-14
#1064 Keep implicit hydrogens information in KET-format

even1024:


core/indigo-core/molecule/src/molecule_json_saver.cpp
```diff
@@ -748,6 +748,14 @@ void MoleculeJsonSaver::saveAtoms(BaseMolecule& mol, JsonWriter& writer)
                     writer.Int(hcount);
                 }
             }
+            else
+            {
+                if (Molecule::shouldWriteHCount(mol.asMolecule(), i) && hcount > 0)
+                {
+                    writer.Key("hImpCount");
```
AlexanderSavelyev, line 755: rename to implicitHCount

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/molecule_json_saver.cpp
even1024: renamed

Issue #1126, opened 2023-05-04
Segfault when iterating CDX file from USPTO downloads


eloyfelix:
**Steps to Reproduce**
Indigo 1.10.0 installed via pip install epam.indigo 
Python 3.10.10
Ubuntu linux

```python
from indigo import Indigo

indigo = Indigo()

for item in indigo.iterateCDXFile("US06174985-20010116-C00003.CDX"):
    print(item.molfile())
```

```

  -INDIGO-05042321392D

 15 14  0  0  0  0  0  0  0  0999 V2000
    0.5133    0.3458    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.5133    0.9048    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0495    1.0778    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
    1.3776    0.6262    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0495    0.1728    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
    1.9421    0.6262    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.2243    0.1375    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.2243    1.1148    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0495    1.6422    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.6140    1.6422    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.4851    1.6422    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    2.5929    1.6581    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.8751    1.1694    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.8751    2.1467    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  2  0  0  0  0
  2  3  1  0  0  0  0
  3  4  1  0  0  0  0
  4  5  2  0  0  0  0
  1  5  1  0  0  0  0
  4  6  1  0  0  0  0
  6  7  1  0  0  0  0
  6  8  1  0  0  0  0
  3  9  1  0  0  0  0
[...]

PR #1130, opened 2023-05-17
#1126 Segfault when iterating CDX file from USPTO downloads

even1024:


utils/indigo-depict/main.c
```diff
@@ -396,8 +396,6 @@ int parseParams(Params* p, int argc, char* argv[])
     indigoSetOptionBool("render-coloring", 1);
     indigoSetOptionBool("render-highlight-color-enabled", 1);
     indigoSetOption("render-superatom-mode", "collapse");
-    indigoSetOption("render-atom-ids-visible", "true");
```
AlexanderSavelyev, line 399: do we need to change depict?

AlexanderSavelyev (approved):

PR #1131, opened 2023-05-17
#1102 When pasting Extended SMILES structure with stereochemistry there are two &1 centers instead of an ABS and an &1

Wadym:


core/indigo-core/molecule/src/smiles_loader.cpp
```diff
@@ -3129,7 +3134,7 @@ void SmilesLoader::_readAtom(Array<char>& atom_str, bool first_in_brackets, _Ato
                 subatom = std::make_unique<QueryMolecule::Atom>(QueryMolecule::ATOM_NUMBER, element);
             else
             {
-                if (element_assigned)
+                if (element_assigned && !stereochemistry_options.ignore_errors)
```
AlexanderSavelyev, line 3137: please add python test for this change

bingo/oracle/src/oracle/ora_wrap.cpp
```diff
@@ -599,7 +599,10 @@ bool OracleStatement::executeSingleInt(int& result, OracleEnv& env, const char*
     statement.prepare();
     statement.defineIntByPos(1, &result);
     if (!statement.executeAllowNoData())
+    {
```
AlexanderSavelyev, line 602: move this change to separate PR

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/smiles_loader.h
```diff
@@ -187,6 +190,7 @@ namespace indigo
 
     private:
         SmilesLoader(const SmilesLoader&); // no implicit copy
+        std::unordered_set<int> _overtly_defined_abs;
```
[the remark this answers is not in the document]
Wadym: As we have discussed we are not going to track changes attributes during SMILES parsing, hence the container is going to be moved into the method.

core/indigo-core/molecule/src/smiles_loader.cpp
Wadym: As per company policy, bugs unrelated to the issue should not be included in the PR. The commit is going to be reverted

AlexanderSavelyev (approved):

Issue #962, opened 2023-01-03
CDXML import: 'superscript' and 'subscript' appears below the letter


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Write 'X' letter, add '2' number with superscript or subscript
3. Save as CDXML or copy from Clipboard
4. Open/Paste from Clipboard saved file
5. Look to text

**Expected behavior**
When you open a saved file with 'superscript' and 'subscript' it is displayed correctly as saved

**Actual behavior**
'superscript' and 'subscript' are not displayed correctly

**Attachments**

https://user-images.githubusercontent.com/62464241/210390732-96152c34-2fef-4e60-bea3-925f1557b2d7.mp4

**Indigo/Bingo version**  
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.9.0-rc.2.0-g40c4e3793-x86_64-linux-gnu-11.2.1

vanoprenko:
This is CDXML import problem. Appears if there is more than one style in single line of text.

even1024:
Fixed in rc3

Zhirnoff:
Bug is reproduced in Indigo 
Version 1.10.0-rc.2

https://user-images.githubusercontent.com/62464241/218065203-3d119df5-e404-4730-adf9-09aab54fd347.mp4

PR #1149, opened 2023-05-31
#962 CDXML import: 'superscript' and 'subscript' is not displayed correctly

Wadym:


even1024 (approved):

core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
```diff
@@ -1367,6 +1367,12 @@ void MoleculeCdxmlSaver::addMetaObject(const MetaObject& obj, int id)
                 s->SetAttribute("font", 4);
                 s->SetAttribute("size", font_size / kCDXMLFonsSizeMultiplier);
                 s->SetAttribute("face", font_face.face);
+                if (font_face.is_superscript)
+                {
+                    s->SetAttribute("face", "64");
```
AlexanderSavelyev, line 1372: move  "64" to constant

core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
```diff
@@ -1367,6 +1367,12 @@ void MoleculeCdxmlSaver::addMetaObject(const MetaObject& obj, int id)
...
+                }
+                if (font_face.is_subscript)
+                    s->SetAttribute("face", "32");
```
AlexanderSavelyev, line 1375: move "32" to constant

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/molecule_cdxml_saver.cpp
Wadym: Added two constans KCDXMLFontStyleSuperscript and KCDXMLFontStyleSubscript. The constants in the CDXML format are described as font style with names subscript and superscript of UINT16 type, so I have taken that into concideration declaring these variables in Indigo

AlexanderSavelyev (approved):

PR #1150, opened 2023-06-01
Added header files to CMake targets

kbelash:
Currently, when you use Visual Studio for development, you may notice that only source files added to C++ projects, but header files are ignored. So, for each C++ project in VS, there are "Source Files" sections with sources, but no "Header Files" with headers. Header files added implicitly as external dependencies -- to "External Dependencies" section. 

This is not very convenient for development process, for example:
* when you want to explore some interfaces in headers, or
* to see, what header files this particular project has, 

"External Dependencies" section can contain dozens of files from different projects and even OS headers.

This PR fixes this problem, adding corresponding header files to all CMake C++ targets.


even1024 (approved):

even1024:
ticket created

even1024:
@kbelash , thanks for your contribution! :-)

Issue #1026, opened 2023-02-07
SVG/PNG: Contracted 'Functional Groups' and 'Salts and Solvents' are rendered expanded when saved


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Put any 'Functional Groups' and 'Salts and Solvents' on canvas (contracted)
3. Save as SVG/PNG
4. Look at rendered preview

**Expected behavior**
Contracted 'Functional Groups' and 'Salts and Solvents' are rendered contracted in preview
**Actual behavior**
Contracted 'Functional Groups' and 'Salts and Solvents' are rendered expanded in preview

**Attachments**
![2023-02-08_01h51_23](https://user-images.githubusercontent.com/62464241/217385259-34392235-2479-417a-96b6-6feeae523705.png)
![2023-02-08_01h51_41](https://user-images.githubusercontent.com/62464241/217385210-c03faa5a-635a-4f95-a0db-6c10fc5919a6.png)
![2023-02-08_01h52_03](https://user-images.githubusercontent.com/62464241/217385212-665aa290-04a9-463f-a937-d1830d8fe84a.png)

**Indigo/Bingo version**  
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.10.0-rc.2.0

PR #1154, opened 2023-06-06
#1026 SVG/PNG: Contracted 'Functional Groups' and 'Salts and Solvents' are rendered expanded when saved

Wadym:


core/render2d/src/render_internal.cpp
```diff
@@ -219,9 +218,24 @@ void MoleculeRenderInternal::setMolecule(BaseMolecule* mol)
...
+    auto isThereAtLeastOneContracted = false;
+    if (superatoms || mulsgroups)
     {
-        _prepareSGroups();
+        auto count = _mol->sgroups.getSGroupCount(SGroup::SG_TYPE_SUP);
+        BaseMolecule& mol = *_mol;
+        for (int i = mol.sgroups.begin(); i != mol.sgroups.end(); i = mol.sgroups.next(i))
+        {
+            SGroup& sgroup = mol.sgroups.getSGroup(i);
+            if (abs(sgroup.contracted) == 1)
```
AlexanderSavelyev, line 229: what is "1" is it a constant?

core/render2d/src/render_internal.cpp
```diff
@@ -867,90 +881,93 @@ void MoleculeRenderInternal::_cloneAndFillMappings()
...
 
     BaseMolecule& mol = *_mol;
-    if (_opt.collapseSuperatoms)
+    if (collapseAtLeastOneSuperatom)
     {
         for (int i = mol.sgroups.begin(); i != mol.sgroups.end(); i = mol.sgroups.next(i))
         {
             SGroup& sgroup = mol.sgroups.getSGroup(i);
-            if (sgroup.sgroup_type == SGroup::SG_TYPE_SUP)
+            if (abs(sgroup.contracted) == 1)
```
AlexanderSavelyev, line 894: what is "1"

core/render2d/src/render_internal.cpp
```diff
@@ -867,90 +881,93 @@ void MoleculeRenderInternal::_cloneAndFillMappings()
...
-                int said = -1;
+                    const Superatom& group = (Superatom&)sgroup;
+                    Vec3f centre;
+                    for (int i = 0; i < group.atoms.size(); ++i)
+                    {
+                        int aid = group.atoms[i];
+                        centre.add(mol.getAtomXyz(aid));
+                    }
+                    centre.scale(1.0f / group.atoms.size());
+                    int said = -1;
```
AlexanderSavelyev, line 906: rename variable to user friendly

AlexanderSavelyev (changes requested):

core/render2d/src/render_internal.cpp
Wadym: The "said" variable was introduced earlier without any comments. I have just rearanged the code. From the code, I assume that "said" stands for Start Atom ID of the bond. There is also "naid" that could mean eNd Atom ID of the bond, and there is "aid" which could mean just Aton ID as it gets index of an atom.
If there are no objections against this assumption I propose to name them as aid : atomID; naid : endAtomID; said : startAtomID

Wadym (commented):
all requested changes have been implemented

core/indigo-core/molecule/molecule.h
```diff
@@ -19,7 +19,9 @@
 #ifndef __molecule_h__
 #define __molecule_h__
 
+#include "molecule/CDXConstants.h"
```
AlexanderSavelyev, line 22: remove CDX constants from molecule.h

core/indigo-core/molecule/multiple_cdx_loader.h
```diff
@@ -23,6 +23,10 @@
 #include "base_cpp/reusable_obj_array.h"
 #include "base_cpp/tlscont.h"
 
+typedef unsigned short int UINT16;
```
AlexanderSavelyev, line 26: do we need it in two places? (CDX constants)?

core/indigo-core/molecule/CDXConstants.h
```diff
@@ -24,6 +24,10 @@
 #endif //  defined( _X86_ ) && !defined( _CS3D )
 #endif // QUADCONST
 
+typedef unsigned short int UINT16;
```
AlexanderSavelyev, line 27: it is worth to move this into common/base_c/defs.h

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/CDXConstants.h
Wadym: Yes it is actually worth, they are rather of common usage than specific to the CDX, at least in their intentions. I have moved all three of them to the common/base_c/defs.h
typedef unsigned short int UINT16;
typedef int INT32;
typedef unsigned int UINT32;

core/indigo-core/molecule/multiple_cdx_loader.h
Wadym: deleted. (This definition was moved here because of a build failure and UINT16 was in error messages as undefined.)

PR #1165, opened 2023-06-30
#1061 Cannot create custom index in python bingo-elastic

### LINKED ISSUE #1061: [Bingo-Elastic] Cannot create custom index in python bingo-elastic
**Description:**
I want to create a custom index for bingo-Elastic without using IndexName to implement library logic for reference API.
The problems I have faced include:
- In Python, we cannot enhance/extend Enum. Therefore, I used a workaround by using the aenum library to extend the IndexName enum with custom name logic;
- The next problem is located at line 147 in the elastic.py file. In that case, I want to index a molecule with a custom index name, but I cannot do so because the index value should be directly from the IndexName enum (see attachments).

**Indigo/Bingo version**  
bingo version: 1.9.0

**Expected behavior**
IndexName should be more flexible to have ability set custom value. 

**Actual behavior**
We should set up Index name only from IndexName enum.

**Attachments**
image

bodyangug:
closes #1061


bingo/bingo-elastic/java/pom.xml
```diff
@@ -53,7 +53,7 @@
         <dependency>
             <groupId>org.elasticsearch.client</groupId>
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
-            <version>7.16.2</version>
```
AlexanderSavelyev, line 56: move java changes into a separate PR

AlexanderSavelyev (changes requested):

bingo/bingo-elastic/java/pom.xml
MysterionRise: agreed

MysterionRise:
@bodyangug I think Java/SSL changes were fixed already - https://github.com/epam/Indigo/commit/9508247149d1b1dbf455e4cb9340566e8d2d9758 - could you doublcheck?

bodyangug:
> @bodyangug I think Java/SSL changes were fixed already - [9508247](https://github.com/epam/Indigo/commit/9508247149d1b1dbf455e4cb9340566e8d2d9758) - could you doublcheck?

I apologize for my mistake. I have checked, and the commit was cleaned up.

bingo/bingo-elastic/java/pom.xml
bodyangug: That issue was resolved at [9508247](https://github.com/epam/Indigo/commit/9508247149d1b1dbf455e4cb9340566e8d2d9758)

AlexanderSavelyev (approved):

PR #1190, opened 2023-07-19
#1191 search molecule using custom index name

bodyangug:
reopen, cause one isuue was found while trying to search at elastic with custom index was recieved error: "Unkown index".
[Issue](https://github.com/epam/Indigo/issues/1191)


MysterionRise:
@bodyangug how we would distinguish between “supported” index and any other random index?

bingo/bingo-elastic/python/bingo_elastic/elastic.py
```diff
@@ -62,11 +62,9 @@ def get_index_name(record: IndigoRecord) -> IndexName:
 def get_record_by_index(
     response: Dict, index: str
 ) -> Union[IndigoRecordMolecule, IndigoRecordReaction]:
-    if index == IndexName.BINGO_MOLECULE.value:
-        return IndigoRecordMolecule(elastic_response=response)
     if index == IndexName.BINGO_REACTION.value:
         return IndigoRecordReaction(elastic_response=response)
-    raise AttributeError(f"Unknown index {index}")
+    return IndigoRecordMolecule(elastic_response=response)
```
MysterionRise, line 67: should we raise error if index isn’t really exists?

bodyangug: That's really hard to track "our" index and other random values because "our" index can be any value that was set up as library-name.

bodyangug: I have provided some solution, check please.

MysterionRise (approved):
LGTM

AlexanderSavelyev (approved):

Issue #1037, opened 2023-02-15
Indigo Toolkit information in the 'About' section that is not relevant for the users


Zhirnoff:
**Background**
At the moment, Ketcher displays Indigo Toolkit information in the 'About' section that is not relevant for the users.
![2023-02-01_01h42_16](https://user-images.githubusercontent.com/62464241/215900521-f73aaba1-b70b-4723-a431-18c17d616afa.png)

**Solution**
We need to hide this information from the 'About' section

Add function to all the APIs (python, java, .Net, WASM)
string versionInfo()

Function should return JSON string
{
    "majorVersion": "1.13"
    "minorVersion":"0",
    "devTag": "rc.1.0",
    "commitHash": "g8b04633b1",
    "compilerVersion": "clang-13.0.0"
    "complierPlatform": "wasm32-wasm"
}

currently version returns "1.13.0-rc.1.0-g8b04633b1-wasm32-wasm-clang-13.0.0"

**Alternatives**

**Additional context**
Ketcher issue: https://github.com/epam/ketcher/issues/2162

@even1024  to update the ticket with API functions requirements

PR #1192, opened 2023-07-21
#1037 Indigo Toolkit information in the 'About' section that is not relevant for the users

dmalysh:


api/c/indigo/src/indigo.cpp
```diff
@@ -41,6 +42,15 @@
 
 static _SessionLocalContainer<Indigo> indigo_self;
 
+static const std::string kVersionInfo = R"({
```
dmalysh, line 45: Not sure about the place where we need to store this data

AlexanderSavelyev: please find where the version is created. The string is populated automatically - but not static

AlexanderSavelyev (changes requested):

api/c/indigo/src/indigo.cpp
dmalysh: Updated. Current output result is number of version: 1.14.0

api/c/indigo/src/indigo.cpp
```diff
@@ -51,6 +51,39 @@ CEXPORT const char* indigoVersion()
...
+        std::string version = indigoVersion();
+
+        const std::string digits = "0123456789";
+        const auto parse = [&](const auto& shift) { return version.substr(shift, version.find_first_of("-\\")); };
+        const auto slice = [&](const auto& value) {
+            version.erase(0, value.size() + 1);
+            return value;
+        };
+
+        const std::string major_version = slice(version.substr(0, parse(1).find_last_of(digits)));
```
AliaksandrDziarkach, line 67: This will cause error if minor version consists of more than one digit

AliaksandrDziarkach:
I think this task can be done in more straight and clear way:
1) Fix api/indigo-version.cmake to set variables for majorVersion, minorVersion, devTag, commitHash
2) Fix api/c/indigo/src/indigo_version.h.in to define corresponding constants and to define compilerVersion and compilerPlatform.
3) Add function which combine all this defines into json.

api/c/indigo/src/indigo.cpp
dmalysh: Fixed.

AliaksandrDziarkach:
> > 3\.
> 
> My previous decision was exactly what you proposed, but I got feedback about I shouldn't change *.cmake files and shouldn't add new constants so I have done it through C++ parse string.

Anyway its was wrong commit - git tag value was not parsed into version vars.

even1024 (approved):

even1024 (approved):

even1024 (approved):

Issue #1161, opened 2023-06-26
Bingo unable to index on SQL Server


boglet:
**Steps to Reproduce**
1. Use Indigo library (Bingo cartridge). Latest release on SQL Server 16.0.1000.6
2. exec bingo.CreateMoleculeIndex 'mycsdb', 'id', 'smiles';

**Expected behavior**
Index expected.

**Actual behavior**
Indexing fails with missing ".resources" manifest (see below)

**Attachments**
```
Msg 6522, Level 16, State 1, Procedure bingo.z_CreateMoleculeIndex, Line 0 [Batch Start Line 0]
A .NET Framework error occurred during execution of user-defined routine or aggregate "z_CreateMoleculeIndex": 
System.Resources.MissingManifestResourceException: Could not find any resources appropriate for the specified culture or the neutral culture.  Make sure "  .resources" was correctly embedded or linked into assembly "bingo-sqlserver" at compile time, or that all the satellite assemblies required are loadable and fully signed.
System.Resources.MissingManifestResourceException: 
   at indigo.Bingo._ExecuteBingoOperation(SqlString bingo_schema, bingoOperationDelegate operationDelegate, bingoGetIndexDataDelegate getBingoDataDelegate, BingoOp op_flags)
   at indigo.Bingo._ExecuteBingoOperationChangeIndex(SqlString bingo_schema, bingoOperationDelegate operationDelegate, bingoGetIndexDataDelegate getBingoDataDelegate, BingoOp flags)
   at indigo.Bingo.CreateIndex(SqlString table, SqlString id_column, SqlString data_column, SqlString bingo_schema, Boolean reaction)
```

**Indigo/Bingo version**  
Downloaded 26/06/2023

**Additional context**

santiagomestrelab:
I have detected the same problem.
Using SQLServerDeveloper version 16.0.1000.6 and BingoDB version bingo-sqlserver-windows-x86_64-1.9.1.587,
and SQLServerExpress version 15.0.4316.3 and BingoDB version bingo-sqlserver-windows-x86_64-1.9.1.544

Regards
Santiago

sapiosciences:
The old 1.9.1 version is working and the binaries released since then do not for MS-SQL.
I would also note the bingo-sqlserver.dll file size is 10MB in 1.9.1, which is twice as what we have in master.

I'm guessing it's missing some assemblies resources.

AlexanderSavelyev:
Hi all

[...]

PR #1204, opened 2023-07-31
 #1161 Bingo unable to index on SQL Server

AliaksandrDziarkach:
Fix CMakeLists.txt and BingoCore.cs to use "indigo" root namespace.
 Regenerate Resource.Designer.cs to use updated namespace.


even1024 (approved):

mjcass:
Hi , I think I am seeing this issue with INDEX not seeing to "stick" - ie I create an index and the tables all seem to be created and immediately after creating it the substructure search is very fast. But connecting to the server the next day the search becomes very slow and it doesn't look like the index is being used. I assume this fix needs to be applied what I am not clear on is how to apply the update? do I need to recompile the program?

Issue #1205, opened 2023-07-31
Reagent located at the bottom of the arrow when opening the RXN V2000 and V3000 files are located on top of the arrow


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Add file on canvas [ketcher (80).zip](https://github.com/epam/Indigo/files/12217625/ketcher.80.zip)
3. Save it as RXN V2000 and RXNV3000 format
4. Open it and look at reagent

**Expected behavior**
Reagent located at the bottom of the arrow as it was saved

**Actual behavior**
Reagent located at the top of the arrow

**Attachments**
 

https://github.com/epam/Indigo/assets/62464241/336831cb-d1ed-49f7-8b06-b355f9f74d97

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Indigo/Bingo version** 
Version 1.13.0-rc.1

PR #1216, opened 2023-08-08
 #1205 Fix reaction json saver layout calculation

AliaksandrDziarkach:
Fix calculation of boundig boxes for reactants and products.


AliaksandrDziarkach:
#1205

even1024 (approved):

Issue #1230, opened 2023-08-17
Unable to save file if the canvas has a reaction arrow and a Functional Group


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Add contracted Functional Group and reaction arrow on canvas
3. Press 'Save as...' button

**Actual behavior**
After clicking the 'Save as...' button, an error message appears: "Convert error! Array: invalid index 1 (size=1)."

**Expected behavior**
File is saved without any errors.

**Screenshots**
![2023-08-17_13h35_02](https://github.com/epam/ketcher/assets/62464241/fbd80925-9195-49d4-8337-dad8d980dad4)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Indigor version** 
[Version 1.14.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)

Zhirnoff:
Task on Ketcher side https://github.com/epam/ketcher/issues/3130

Zhirnoff:
It is also not possible to save the attached file. A similar error appears.
[Multi (3).zip](https://github.com/epam/Indigo/files/12370303/Multi.3.zip)
![2023-08-17_17h09_00](https://github.com/epam/Indigo/assets/62464241/84a717a2-f1b0-4ee8-b7f7-e6e9523bb637)

PR #1238, opened 2023-08-21
Backmerge: #1230 Unable to save file if the canvas has a reaction arrow and a Fu…

even1024:
…nctional Group (#1237)


AliaksandrDziarkach (approved):
LGTM

PR #1248, opened 2023-08-25
yqiao/master

sapiosciences-dev:
For Aleksandr Savelev as requested.

### DIFF
--- a/bingo/bingo-elastic/python/bingo_elastic/queries.py
+++ b/bingo/bingo-elastic/python/bingo_elastic/queries.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional
 
 from indigo import Indigo, IndigoObject  # type: ignore
 
-from bingo_elastic.model.record import IndigoRecord, IndigoRecordMolecule
+from bingo_elastic.model.record import IndigoRecord, IndigoRecordMolecule, IndigoRecordReaction
 from bingo_elastic.utils import PostprocessType, head_by_path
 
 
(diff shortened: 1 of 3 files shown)


sapiosciences-dev:
For aromotize() call on registration and post-processing perhaps it can preserve backward compatibility should we have a toggle on the ElasticRepository and AsyncElasticEpiository class with a default value set to True (so it can be set to false manually for legacy connections)?

bingo/bingo-elastic/python/bingo_elastic/model/record.py
```diff
@@ -108,6 +109,12 @@ class IndigoRecord:
     elastic_response = WithElasticResponse()
     record_id: Optional[str] = None
     error_handler: Optional[Callable[[object, BaseException], None]] = None
+    # Sort for page cursor
+    _sort: list
```
khyurri, line 113: Seems that `IndigoRecord` not the right place for storing sort cursor. This is representation of IndigoObject. Should we move `_sort` to other place?

bingo/bingo-elastic/python/bingo_elastic/queries.py
```diff
@@ -47,23 +47,60 @@ def compile(
...
+            bool_head[parent_term] = []
+        bool_head[parent_term].append(
             {"match": {self.field: {"query": self._value, "boost": 0}}}
         )
-        default_script_score(query)
+        if not self.is_substructure:
+            default_script_score(query)
+
+
+class TermQuery(CompilableQuery):
```
khyurri, line 75: Should we add some tests for this Query? We don't have much documentation, but tests can help us to understand usage and use-cases. Also, if this PR will be merged, we need to migrate same functionality into Java wrapper and tests will help us to achieve that

khyurri (changes requested):
Apologies for long review and thank you for your PR. Great to have new Query (TermQuery), but we would clarify some questions.

bingo/bingo-elastic/python/bingo_elastic/model/record.py
sapiosciences-dev: Personal opinion i think since elastic_response = WithElasticResponse() is already in IndigoRecord, _sort can fit here as well. Since _sort is under the context of elastic response. It is used as the cursor of the current elastic search under the elastic_response.

I don't mind an alternative storage location though. What would you suggest?

Issue #1184, opened 2023-07-19
Support monomer templates export into KET-format


even1024:
Background
Indigo macromolecules support

**Solution**
Implement export from Indigo T-Groups into monomerTemplate objects of KET-format.
Tests should cover the conversion of MOL-V3000 templates into KET-monomerTemplate.

PR #1251, opened 2023-09-04
#1184 Support monomer templates export into KET-format

even1024:


core/indigo-core/molecule/src/molecule_json_loader.cpp
```diff
@@ -956,6 +973,204 @@ void MoleculeJsonLoader::parseProperties(const rapidjson::Value& props, BaseMole
...
+                const auto& target_pos = mol.getAtomXyz(target_atom);
+                sa.bonds.push(vx.neiEdge(i));
+                brackets.emplace_back((target_pos.x - src_pos.x) / 2, (target_pos.y - src_pos.y) / 2);
+            }
+        }
+    }
+
+    // fill brackets
+
+    for (int i = 0; i < brackets.size(); i += 2)
```
AliaksandrDziarkach, line 1003: Looks strange.
Is it really each bracket should have second member set to 0 if brackets count is odd?
May be it should be only for last bracket?

AlexanderSavelyev (changes requested):
remove huge tests
add helmClass
change id generation
update class by missing values

core/indigo-core/molecule/src/molecule_json_loader.cpp
even1024: fixed

AlexanderSavelyev (approved):

Issue #1183, opened 2023-07-19
Support monomer templates import from KET-format


even1024:
**Background**
Indigo macromolecules support

**Solution**

Implement monomerTemplate import from KET into Indigo T-Groups.
Tests should cover the conversion of monomerTemplate into MOL-V3000 format.

PR #1259, opened 2023-09-08
#1183 Support monomer templates import from KET-format

even1024:


AlexanderSavelyev (changes requested):
remove huge tests
add helmClass
change id generation
update class by missing values

Issue #1272, opened 2023-09-13
After opening a saved Extended SMILES file Single Up Bond is changed to Single Down Bond


Zhirnoff:
**Steps to Reproduce**
1. Load the following molfile [atrop_enhanced_stereo.zip](https://github.com/epam/Indigo/files/12094011/atrop_enhanced_stereo.zip)
2. Save into Extended SMILES
3. Open saved file

![2023-09-13_16h35_44](https://github.com/epam/Indigo/assets/62464241/76e74c91-263a-403b-a7e5-9291ea275df8)

**Expected behavior**
Stereo information for bond and atom is kept. Single Up Bond is kept.

**Actual behavior**
Stereo information for bond and atom is kept. Single Up Bond is changed to Single Down Bond.

#185522135

even1024:
Fixed in 1.15.0.rc.2

PR #1273, opened 2023-09-14
#1272 After opening a saved Extended SMILES file Single Up Bond is changed to Single Down Bond

even1024:


api/c/indigo/src/indigo_layout.cpp
```diff
@@ -58,6 +58,9 @@ CEXPORT int indigoLayout(int object)
             ml.max_iterations = self.layout_max_iterations;
             ml.bond_length = 1.6f;
             ml.layout_orientation = (layout_orientation_value)self.layout_orientation;
+            bool has_atropisomery = mol->hasAtropisomericCenter();
+            if (has_atropisomery)
+                ml.respect_existing_layout = true;
```
AlexanderSavelyev, line 63: respect_existing_layout is not used anywhere

AlexanderSavelyev (changes requested):
add tests

AlexanderSavelyev (approved):

Issue #1277, opened 2023-09-14
Atropisomer centers support


even1024:
Implement specific type of stereocenter, keep non-standard bond directions inside the atropoisomeric center.  
Add bond direction restore after the layout.

PR #1280, opened 2023-09-15
#1277 SMILES with bad stereochemistry is not parsed

even1024:


AlexanderSavelyev (changes requested):
move forced stereo bonds into atropoisomers centers

Issue #1285, opened 2023-09-21
Add new query properties support to KET format


AliaksandrDziarkach:
Add json read/write support for next atom query properties:

```json
        "queryProperties": {
          "type": "object",
          "properties": {
            "aromaticity": {
              "type": "string",
              "enum": ["aromatic", "aliphatic"]
            },
            "degree": {
              "type": "integer",
              "minimum": 0,
              "maximum": 6
            },
            "ringMembership": {
              "type": "integer",
              "minimum": 0,
              "maximum": 5
            },
            "ringSize": {
              "type": "integer",
              "minimum": 0,
              "maximum": 5
            },
            "connectivity": {
              "type": "integer",
              "minimum": 0,
              "maximum": 5
            },
            "ringConnectivity": {
              "type": "integer",
              "minimum": 0,
              "maximum": 5
            },
            "atomicMass": {
              "type": "integer",
              "minimum": 0,
              "maximum": 1000
            },
            "chirality": {
              "type": "string",
              "enum": ["clockwise", "anticlockwise"]
            },
            "customQuery": {
              "type": "string"
            }
          }
```

PR #1286, opened 2023-09-26
#1285 Add new query properties support to KET format

AliaksandrDziarkach:
Add query properties load/save support to ket format.
Add UTs.


api/tests/integration/test.py
AliaksandrDziarkach, line 1: Strings moved to constants.
Added code to treat [new] UT as failed to prevent forgotten UT.

even1024 (approved):

Issue #1332, opened 2023-10-16
Aromatic atom list is not recognized in SMARTS query


urszula-gawlowska:
When Indigo converts a SMARTS query with a list atoms in aromatic format (lowercase atom symbols, e.g. "c,n") into a structure, the list is not recognized properly and is interpreted as a custom query in the returned structure.

For example, the following query results in a molecule with custom query "c,n" instead of atom list with carbon and nitrogen.
`c1-[#6]=[#6]-[#6]=[#6]-[c,n]=1`

vanoprenko:
https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html

vanoprenko:
Causes https://github.com/epam/ketcher/issues/3455

vanoprenko:
**Technical details based on investigation:**

Payload for layout request:

```
{
  "struct": "c1-[#6]=[#6]-[#6]=[#6]-[c,n]=1",
  "output_format": "chemical/x-indigo-ket",
  "options": {
    "smart-layout": true,
    "ignore-stereochemistry-errors": true,
    "mass-skip-error-on-pseudoatoms": false,
    "gross-formula-add-rsites": true,
    "aromatize-skip-superatoms": true,
    "dearomatize-on-load": false,
    "ignore-no-chiral-flag": false
  }
}
```
Response for layout request:

```
{
    "format": "chemical/x-indigo-ket",
    "original_format": "chemical/x-daylight-smiles",
[...]

PR #1367, opened 2023-10-22
 #1332 Aromatic atom list is not recognized in SMARTS query

AliaksandrDziarkach:
Add code to understand lists like (a1&p1&pn)|(an&p1&pn) as (a1|a2)&p1&pn 
Add UT


core/indigo-core/molecule/src/query_molecule.cpp
```diff
@@ -2333,7 +2333,97 @@ QueryMolecule::Atom* QueryMolecule::stripKnownAttrs(QueryMolecule::Atom& qa)
...
+                return false;
+                break;
+            case ATOM_NUMBER:
+                if (child_prop->value_min != child_prop->value_max)
+                    return false;
+                atoms_list.emplace_back(child_prop->value_min);
+                atom_not_found = false;
+                break;
+            default:
+                Atom* copy = new Atom();
```
AlexanderSavelyev, line 2387: use unique ptrs

AlexanderSavelyev (changes requested):

Issue #1293, opened 2023-10-02
Indigo renderer does not support some query options


AliaksandrDziarkach:
Indigo renderer does not support sumo query options, as result query with these options rendered wrong.

Elements always rendered as in periodic table which is wrong for queries.

renderer_internal.cpp should be updated to support all query options and use correct aromatic/aliphatic notation - C,c or `#6` corresponding to query.

PR #1390, opened 2023-11-01
#1293 Indigo renderer does not support some query options 

AliaksandrDziarkach:


core/indigo-core/molecule/src/query_molecule.cpp
```diff
@@ -2292,8 +2302,7 @@ bool QueryMolecule::isKnownAttr(QueryMolecule::Atom& qa)
     return (qa.type == QueryMolecule::ATOM_CHARGE || qa.type == QueryMolecule::ATOM_ISOTOPE || qa.type == QueryMolecule::ATOM_RADICAL ||
             qa.type == QueryMolecule::ATOM_VALENCE || qa.type == QueryMolecule::ATOM_TOTAL_H || qa.type == QueryMolecule::ATOM_SUBSTITUENTS ||
             qa.type == QueryMolecule::ATOM_SUBSTITUENTS_AS_DRAWN || qa.type == QueryMolecule::ATOM_RING_BONDS ||
-            qa.type == QueryMolecule::ATOM_RING_BONDS_AS_DRAWN || qa.type == QueryMolecule::ATOM_UNSATURATION || qa.type == QueryMolecule::ATOM_AROMATICITY ||
```
AliaksandrDziarkach, line 2295: Was changed in previous commits.
But then I realize that this this change not need.
Because it broke render UT - return to original code.

AlexanderSavelyev (approved):

Issue #1395, opened 2023-11-07
For KET monomer templates the fields class and naturalAnalog should be optional  


even1024:
**Steps to Reproduce**
KET-loader won't load monomer templates with missing "class" field.

**Expected behavior**
Make "class"and "naturalAnalog"deducible from "classHELM"and "naturalAnalogShort"if not specidied.

PR #1396, opened 2023-11-07
#1395 Class and naturalAnalog of KET monomerTemplate should be optional

even1024:


core/indigo-core/molecule/src/molecule_json_loader.cpp
```diff
@@ -1115,6 +1115,22 @@ void MoleculeJsonLoader::fillXBondsAndBrackets(Superatom& sa, BaseMolecule& mol)
...
 }
 
+std::string MoleculeJsonLoader::monomerTemplateClass(const rapidjson::Value& monomer_template)
+{
+    std::string result;
+    if (monomer_template.HasMember("class"))
+        result = monomer_template["class"].GetString();
+    else if (monomer_template.HasMember("classHELM"))
+    {
+        result = monomer_template["classHELM"].GetString();
```
AlexanderSavelyev, line 1125: add other HELM classes
add tests for other classes

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/molecule_json_loader.cpp
even1024: tests added

AlexanderSavelyev (approved):

PR #1418, opened 2023-11-22
Bugfix/1415 wrong s group count in mol v3000 header

AliaksandrDziarkach:
Fix molfile saver to add ImplicitH S-Groups before write header


AlexanderSavelyev:
closed as a duplicate #1418

Issue #1423, opened 2023-12-04
Common atoms loaded as aliphatic in SMARTS mode


AliaksandrDziarkach:
Load some common atom in SMILES mode - got exactly this atoms.
Loads same atoms in SMARTS mode - got atoms with 'A' atribute which means "Aliphatic"
For example load "BrI" as SMILES - got "Br-I", loads as SMARTS - got "Br;A-I;A", which is a bug.

According to [documentation](https://www.daylight.com/dayhtml/doc/theory/theory.smiles.html#RTFToC28) "only atoms on the following list can be considered aromatic: C, N, O, P, S, As, Se", so Aliphatic has no sence for all other atoms.

PR #1424, opened 2023-12-06
#1423 Common atoms loaded as aliphatic in SMARTS mode

AliaksandrDziarkach:
Fix SMARTS load and save to correct use aliphatic flag.


api/tests/integration/ref/formats/reaction_to_smarts.py.out
```diff
@@ -1 +1 @@
-[#6]=[#6]-[#35].[#7](-[#1])-[#6]>[#17]>[#6]=[#6]-[#7]-[#6]
+[#6]=[#6]-Br.[#7](-[H])-[#6]>Cl>[#6]=[#6]-[#7]-[#6]
```
AlexanderSavelyev, line 1: why do we remove brackets? is it correct notation?

api/tests/integration/ref/formats/serialize_check.py.out
```diff
@@ -104,7 +104,7 @@ components: 119 119
 data Sgroups: 1 1 
 generic Sgroups: 1 1 
 heavy atoms: 419 419 
-hydrogens: 500 500 
+hydrogens: 503 503
```
AlexanderSavelyev, line 107: why it is changed? is it because the brackets removed?

AlexanderSavelyev (approved):

Issue #1436, opened 2023-12-11
Expose Fold/Unfold hydrogens function in indigo API


AliaksandrDziarkach:
Indigo should allow retrieval of chemical structures with either implicit or explicit representation of hydrogen atoms. For this purpose a new API function should be introduced: 

New API function "convert_explicit_hydrogens":
At least two parameters:
1) struct - molecule data
2) mode - auto/fold/unfold. auto is default. If auto used - should check explicit hydrogens and if any explicit hydrogen defined - do "fold"(which means convert explicit hydrogens to implicit), otherwise - "unfold"(which means convert implicit hydrogens to explicit).

PR #1437, opened 2023-12-12
 #1436 Expose Fold/Unfold hydrogens function in indigo API

AliaksandrDziarkach:
Add function convert_explicit_hydrogens.
Add UT.

### DIFF
--- a/api/c/indigo/src/indigo_layout.cpp
+++ b/api/c/indigo/src/indigo_layout.cpp
@@ -27,7 +27,7 @@
 #include <algorithm>
 #include <vector>
 
-CEXPORT int indigoLayout(int object)
+static int _indigoLayout(int object, bool respect_old)
 {
     INDIGO_BEGIN
     {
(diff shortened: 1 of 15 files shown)


utils/indigo-service/backend/service/v2/indigo_api.py
```diff
@@ -888,6 +888,115 @@ def convert():
         )
 
 
+@indigo_api.route("/convert_explicit_hydrogens", methods=["POST"])
```
AlexanderSavelyev, line 891: please add tests for indigo api

AlexanderSavelyev (changes requested):

api/c/indigo/src/indigo_layout.cpp
```diff
@@ -126,6 +128,16 @@ CEXPORT int indigoLayout(int object)
     INDIGO_END(-1);
 }
 
+CEXPORT int indigoLayout(int object)
+{
+    return _indigoLayout(object, false);
+}
+
+CEXPORT int indigoLayoutRespectOld(int object)
```
AlexanderSavelyev, line 136: create parameter layout-preserve-existing instead of new function

AlexanderSavelyev (changes requested):

api/c/indigo/src/indigo_layout.cpp
AliaksandrDziarkach: Updated

AlexanderSavelyev (approved):

Issue #1522, opened 2024-01-12
A lot of warning when compile  indigo-core-unit-tests


AliaksandrDziarkach:
When compile indigo-core-unit-tests compiler generate a lot of warnings.

PR #1523, opened 2024-01-12
 #1522 A lot of warning when compile indigo-core-unit-tests

AliaksandrDziarkach:
For MSVC add #pragma warning to disable DLLEXPORT related warnings which generated a lot of warnings.
For MSVC add "-WX" option (treat compiler warnings as errors) in projects "indigo-core" and "render2d"
Cleanup "render2d" project to "W4" level and set warning level to W4 for MSVC
Cleanup "indigo-core" project to "W3" level and set warning level to W3 for MSVC

build_indigo_libs_x86_64 (windows) log size reduced from 6400 lines to 2100 lines.

### DIFF
--- a/api/c/indigo/src/option_manager.h
+++ b/api/c/indigo/src/option_manager.h
@@ -19,6 +19,11 @@
 #ifndef __otion_manager_h__
 #define __otion_manager_h__
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4251)
+#endif
+
 #include "base_cpp/os_sync_wrapper.h"
 #include "base_cpp/red_black.h"
 
@@ -185,5 +190,9 @@ protected:
 private:
     IndigoOptionManager(const IndigoOptionManager&);
 };
+
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
 
 #endif //__otion_manager_h__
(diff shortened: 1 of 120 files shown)


core/indigo-core/molecule/src/molecule_arom.cpp
```diff
@@ -641,7 +641,7 @@ QueryMoleculeAromatizer::PiValue QueryMoleculeAromatizer::_getPiLabel(int v_idx)
         }
     }
 
-    if (_options.aromatize_skip_superatoms && _inside_superatoms.size() && _inside_superatoms.find(v_idx) != _inside_superatoms.end())
+    if (_options.aromatize_skip_superatoms && _inside_superatoms.size() && _inside_superatoms.find(v_idx))
```
AliaksandrDziarkach, line 644: "_inside_superatoms.find(v_idx)" returned bool, not int as ".end()"

api/c/indigo/src/option_manager.h
```diff
@@ -19,6 +19,11 @@
 #ifndef __otion_manager_h__
 #define __otion_manager_h__
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4251)
```
AliaksandrDziarkach, line 24: 4251 and 4275 warnings in MSVC is about dllexport class derived from non-dllexport class or containing non-dllexport members.

AlexanderSavelyev (approved):

Issue #1463, opened 2023-12-28
Macro: Some molecules are not perfect on preview tooltip


AlexeyGirin:
**Steps to Reproduce**
1. Switch to Macromolecules view
2. Load this file: [BadR3Position.zip](https://github.com/epam/ketcher/files/13784349/BadR3Position.zip) (unzip first)

3. Hover mouse on both peptides

**Actual behavior**
![image](https://github.com/epam/ketcher/assets/26869421/784cb98b-e0ab-443f-b2c1-d65315d113c9)
![image](https://github.com/epam/ketcher/assets/26869421/bc1cdcdf-4b7a-41c7-a880-6abd778e9e22)

**Expected behavior**
Should look better

**Desktop (please complete the following information):**
 - OS: win10
 - Browser Chrome 120.0.6099.130 (Official Build) (64-bit)

**Ketcher version** [e.g. v2.4.2]. 
Ketcher: Version 2.17.0-rc.2
Build at 2023-12-27; 13:44:29

Indigo Toolkit
Version 1.17.0-rc.3.0-g1a7f60990-wasm32-wasm-clang-12.0.0

Issue #1429, opened 2023-12-08
CPP Header autogeneration for integrated monomer templates


even1024:
**Background**
#1412  Introdoced hardcoded basic monomer templates for phosphates, nucleobases and sugars.
The template data saved here: https://github.com/epam/Indigo/blob/master/core/indigo-core/molecule/monomers_basic_templates.h
This file is inconvenient to edit."

**Solution**
Monomer's templates should be stored in regular .KET file like monomers_basic_templates.ket and corresponding monomers_basic_templates.h should be autogenerated with cmake.

PR #1546, opened 2024-01-15
#1463 Macro: Some molecules are not perfect on preview tooltip

even1024:


core/indigo-core/molecule/monomers_basic_templates.h
```diff
@@ -113,15 +113,15 @@ namespace indigo
                                          "                \"attachmentAtom\": 0,\n"
```
AliaksandrDziarkach, line 113: May be better use raw string for multiline strings?
Something like
char *string =  R""(multi
    "multi": "line"
 string)"";

core/indigo-core/molecule/src/query_molecule.cpp
```diff
@@ -1029,6 +1029,11 @@ const int QueryMolecule::getTemplateAtomTemplateIndex(int idx)
     return -1;
 }
 
+void QueryMolecule::getTemplatesMap(std::unordered_map<std::pair<std::string, std::string>, std::reference_wrapper<TGroup>, pair_hash>& templates_map)
```
AliaksandrDziarkach, line 1032: templates_map unused

core/indigo-core/molecule/monomers_basic_templates.h
even1024: this header will be autogenerated. here is a ticket: https://github.com/epam/Indigo/issues/1429

core/indigo-core/molecule/src/query_molecule.cpp
even1024: fixed

AliaksandrDziarkach (commented):
LGTM

AliaksandrDziarkach (approved):

Issue #1476, opened 2024-01-03
Aromatization/Dearomatization wipes out SOME Ring bond count values


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.2.0-g369c82f8c-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.15.0-rc.4 Build at 2023-10-25; 14:15:43
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1446

https://github.com/epam/Indigo/assets/26869421/eabc080d-f93c-42c0-a767-38dab40179e1

**Describe the bug**
Aromatization/Dearomatization wipes out SOME Ring bond count values

**Steps to Reproduce**
**Steps to Reproduce**
1. Load molecule from file: [RingBondCount.zip](https://github.com/epam/Indigo/files/13822424/RingBondCount.zip)
2. Press Aromatize button (Alt+A)

**Actual behavior**
rb5, rb6, rb7, rb8, rb9 values disappear
(Same behavior with dearomatization)

**Expected behavior**
All values remain in place

AlexeyGirin:
Confirmed as bug

PR #1566, opened 2024-01-16
 #1476 Aromatization/Dearomatization wipes out SOME Ring bond count values

AliaksandrDziarkach:
Fix ket file loader/saver.
Fix molfile saver.
Add UT.


even1024 (approved):
everything is ok

Issue #1478, opened 2024-01-04
Dearomatization causes exception in case of Implicit H count query feature set to 4 (i.e. more than 2)


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.2.0-g369c82f8c-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.15.0-rc.4 Build at 2023-10-25; 14:15:43
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1446

**Describe the bug**
Dearomatization causes exception in case of Implicit H count query feature set to 4 (i.e. more than 2)

**Steps to Reproduce**
1. Load from file: [Implicit H 4 (2).zip](https://github.com/epam/Indigo/files/13828997/Implicit.H.4.2.zip)
2. Press Dearomatize button (Ctrl+Alt+A)

**Actual behavior**
System throws exception:
Error Message
IndigoException: dearomatization: Dearomatization groups: atom_aromatic_connectivity < 0 on C having 6 drawn bonds, charge 0, and 0 radical electrons

**Expected behavior**
Exception don't happen
![image](https://github.com/epam/Indigo/assets/26869421/27b3ca0d-adfe-4a80-a56b-28951e9c3209)

AlexeyGirin:
Confirmed as bug
It must not dearmatize in that case (i.e. 4 and more)

AlexeyGirin:
Not fixed.
Dearomatization simply clean up Implicit H value but not dearomatize the molecule.
More over - it seems that Implicit H value set on already aromatized ring is the different something (it doesn't visible on the canvas, for example)

AliaksandrDziarkach:
> **Expected behavior**
Exception don't happen

Exception issue fixed.
If you found other issues - please feel corresponding bugs.

AlexeyGirin:
> If you found other issues - please fill separate corresponding bugs.
:-)))))

AlexeyGirin:
Fixed.
 - Ketcher Version 2.18.0-rc.4 Build at 2024-02-06; 13:45:28
 - Indigo Toolkit Version 1.18.0-rc.6.0-g98bf384e0-x86_64-linux-gnu-11.2.1

PR #1570, opened 2024-01-16
 #1478 Dearomatization causes exception in case of Implicit H count query feature set to 4 (i.e. more than 2)

AliaksandrDziarkach:
Remove exception. Add UT


even1024 (approved):
everything is ok

Issue #1525, opened 2024-01-13
System attach two explicit hydrogens to aromatized ring


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1436
 
**Describe the bug**
System attach two explicit hydrogens to aromatized ring and layout is wrong by default

**Steps to Reproduce**
1. Load from file: [Add explicit hydrogens works not good.zip](https://github.com/epam/Indigo/files/13927691/Add.explicit.hydrogens.works.not.good.zip)  (unzip first)
2. Press Add/Remove explicit hydrogens button

**Actual behavior**
![image](https://github.com/epam/Indigo/assets/26869421/1e95dbb0-5efb-4295-95e5-52483e575fbc)
System attach two hydrogens and layout is not good

**Expected behavior**
Add explicit hydrogens works fine
![image](https://github.com/epam/Indigo/assets/26869421/810ee750-a635-4dc5-9ab5-f99547fe084a)

AlexeyGirin:
Fixed.
![image](https://github.com/epam/Indigo/assets/26869421/05fe2b17-1668-4e95-ad8f-2312179ed3c0)

PR #1577, opened 2024-01-17
 #1525 System attach two explicit hydrogens to aromatized ring

AliaksandrDziarkach:
Fix connection calculation for aromatic bonds. Add UT


even1024 (approved):
looks ok

Issue #1573, opened 2024-01-16
Add/Remove explicit hydrogens can't be applied to fullerene C60 - system throws exception


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1436
 
**Describe the bug**
Add/Remove explicit hydrogens can't be applied to fullerene C60 - system throws exception

**Steps to Reproduce**
1. Load from file: [Add-Remove explicit hydrogens can't be applied to fullerene C60 - system throws exception.zip](https://github.com/epam/Indigo/files/13955866/Add-Remove.explicit.hydrogens.can.t.be.applied.to.fullerene.C60.-.system.throws.exception.zip) (unzip first)
2. Press Add/Remove explicit hydrogens button
![image](https://github.com/epam/Indigo/assets/26869421/2d5bb479-c251-40d3-bf6a-4d2f82479d39)

**Actual behavior**
System throws exception: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
![image](https://github.com/epam/Indigo/assets/26869421/90168ad1-df0b-444d-89d0-44ec18b14805)
![image](https://github.com/epam/Indigo/assets/26869421/3f7baf04-b254-4bd1-81af-72595f250a07)
[uat.test.lifescience.opensource.epam.com-1705440108477.log](https://github.com/epam/Indigo/files/13955874/uat.test.lifescience.opensource.epam.com-1705440108477.log)

**Expected behavior**
No exception
![image](https://github.com/epam/Indigo/assets/26869421/2d5bb479-c251-40d3-bf6a-4d2f82479d39)

AlexeyGirin:
Fixed.
Version 1.18.0-rc.5.0-gb21ab5ad5-wasm32-wasm-clang-12.0.0

PR #1585, opened 2024-01-17
  #1573 Add/Remove explicit hydrogens can't be applied to fullerene C60 - system throws exception

AliaksandrDziarkach:
Add code to skip layout if no hydrogens added. Add UT.


even1024 (approved):
everything ok

Issue #1589, opened 2024-01-18
Apply hydrogens folding/unfolding with respect to selected atoms 


vanoprenko:
Hydrogens folding/unfolding implemented in scope of https://github.com/epam/Indigo/issues/1436 is currently applied to all atoms in the destination structure. Enhance this functionality, so that folding/unfolding is applied to selected atoms only. If no atoms are selected, the whole structure should be affected.
KET format should be enhanced in scope of this task, so that selection flag is specified for each selectable entity (atom, bond, static drawing, etc).

AliaksandrDziarkach:
Backmerge https://github.com/epam/Indigo/pull/1655

PR #1590, opened 2024-01-18
 #1589 Apply hydrogens folding/unfolding with respect to selected atoms

AliaksandrDziarkach:
Add "selected" support. Add UT.


even1024 (approved):
looks ok

Issue #1598, opened 2024-01-22
Macro: V3000 export: leaving groups are displayed as side chain connections for standard presets added to canvas


Zhirnoff:
**Steps to Reproduce**
1. Go to 'Macromolecules'
2. Add several RNA presets on canvas (for ex. A, C, G) and connect them with bonds.
3. Remove the last phosphate
4. Go to 'Open structure' -> 'Paste from clipboard' (AG: Seems it is mol v3000 way)
5. Type RNA:ACG
6. Save to mol file
7. Open this file in another sketcher 

**Actual behavior**
Sequences look different.
The one, that was created on canvas, has explicit living groups connected to the main sequence. 
Chemical structure is correct for both sequences.

**Expected behavior**
Sequences should look the same. Leaving groups should be hidden inside Sgroup.

**Screenshots**
![image](https://github.com/epam/ketcher/assets/143799055/ac71d613-0a06-4587-a1fe-f08f8d7aae7d)

[Sequence_vs_preset_on_canvas_export_v3000.zip](https://github.com/epam/ketcher/files/13991003/Sequence_vs_preset_on_canvas_export_v3000.zip)
[ketcher (24).zip](https://github.com/epam/ketcher/files/14008452/ketcher.24.zip)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro
 - Browser Chrome
 - Version 120.0.6099.130 (Official Build) (64-bit)

**Ketcher version** [e.g. v2.4.2]. 
[Version 2.18.0-rc.2](https://lifescience.opensource.epam.com/ketcher/index.html)
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.18.0-rc.3.0-ge65b5cde3-wasm32-wasm-clang-12.0.0

Zhirnoff:
@even1024 When exporting a single chain on canvas to external editor, everything appears normal, and hydrogens are not shown. However, when exporting two chains, they overlap, and one of the chains displays hydrogens. Additionally, when saving in mol V3000 format, SEQID is not assigned to the second structure.
![2024-02-01_20h51_43](https://github.com/epam/Indigo/assets/62464241/88dfced2-5526-47e8-858f-e89cbde8b0a6)

https://github.com/epam/Indigo/assets/62464241/209afe71-a85a-4eb5-b720-9a7d73c93de2

File saved in Ketcher: [structure.zip](https://github.com/epam/Indigo/files/14139572/structure.2.zip)
[...]

PR #1606, opened 2024-01-25
#1598 Macro: V3000 export: leaving groups are displayed as side chain connections for standard presets added to canvas

even1024:


AliaksandrDziarkach (approved):
LGTM

Issue #1533, opened 2024-01-13
System attach two explicit hydrogens to atoms connected to "any type" bonds


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1436
 
**Describe the bug**
System attach two explicit hydrogens to atoms connected to "any type" bonds

**Steps to Reproduce**
1. Load from file: [AnyTypeBonds.zip](https://github.com/epam/Indigo/files/13929516/AnyTypeBonds.zip) (unzip first)
2. Press Add/Remove explicit hydrogens button

**Actual behavior**
![image](https://github.com/epam/Indigo/assets/26869421/117c8fff-5515-400c-a028-691e90fba873)
System attach two hydrogens to atoms connected to "any type" bonds

**Expected behavior**
No connection happens
![image](https://github.com/epam/Indigo/assets/26869421/19736ead-5fe9-41a6-a4f1-c7ed64e6b358)

AlexeyGirin:
As per @AlexanderSavelyev 
Any uncertain bond - should not be added by hydrogens

Fix in Query Features - Phase II

AlexeyGirin:
Fixed.
![image](https://github.com/epam/Indigo/assets/26869421/79afca1c-6d3e-4dfe-a864-0e6af1770cd6)

PR #1612, opened 2024-01-25
 #1533 System attach two explicit hydrogens to atoms connected to "any type" bonds

AliaksandrDziarkach:
Fix code. Add UT


even1024 (approved):
everything ok

Issue #1538, opened 2024-01-13
Add/Remove explicit hydrogens feature doesn't work if atom with problem valence present on canvas (crash happens)


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1436
 
**Describe the bug**
Add/Remove explicit hydrogens feature doesn't work if atom with problem valence present on canvas (crash happens)

**Steps to Reproduce**
1. Load from file: [BadValenceAtomCausesCrash.zip](https://github.com/epam/Indigo/files/13929604/BadValenceAtomCausesCrash.zip) (unzup first)
2. Press Add/Remove explicit hydrogens button

**Actual behavior**
System throws exception
IndigoException: element: bad valence on C having 5 drawn bonds, charge 0, and 0 radical electrons
[uat.test.lifescience.opensource.epam.com-1705183134610.log](https://github.com/epam/Indigo/files/13929607/uat.test.lifescience.opensource.epam.com-1705183134610.log)

**Expected behavior**
Add/Remove explicit hydrogens feature works as intended
![image](https://github.com/epam/Indigo/assets/26869421/5b8e60d6-bf5f-4be3-b1fb-0074fb13f03a)

AlexeyGirin:
Fixed.
Version 1.18.0-rc.5.0-gb21ab5ad5-wasm32-wasm-clang-12.0.0
![image](https://github.com/epam/Indigo/assets/26869421/da234c14-6393-413e-b336-3149ab6aaa31)

PR #1615, opened 2024-01-25
 #1538 Add/Remove explicit hydrogens feature doesn't work if atom with problem valence present on canvas (crash happens)

AliaksandrDziarkach:
Fix code. Add UT


even1024 (approved):
look ok

Issue #1536, opened 2024-01-13
Add/Remove explicit hydrogens feature doesn't work for reactions on canvas (crash happens)


AlexeyGirin:
**Environment details  (please complete the following information):**
 - Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0
 - Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
 - Chrome Version 120.0.6099.130 (Official Build) (64-bit)
 - Win10
 - Issue found while testing https://github.com/epam/Indigo/issues/1436
 
**Describe the bug**
Add/Remove explicit hydrogens feature doesn't work for reactions on canvas

**Steps to Reproduce**
1. Load from file: [ReactionCrash.zip](https://github.com/epam/Indigo/files/13929560/ReactionCrash.zip) (unzup first)
2. Press Add/Remove explicit hydrogens button

**Actual behavior**
System throws exception
IndigoException: core: <reaction> is not a base molecule
[uat.test.lifescience.opensource.epam.com-1705181853119.log](https://github.com/epam/Indigo/files/13929561/uat.test.lifescience.opensource.epam.com-1705181853119.log)

**Expected behavior**
Add/Remove explicit hydrogens feature works as intended
![image](https://github.com/epam/Indigo/assets/26869421/58ba7e5a-907c-4aed-bfc3-113077250ffc)

AliaksandrDziarkach:
Should be fixed in 1.18.0-rc.4

AlexeyGirin:
Already fixed in 1.18.0-rc.4

AlexeyGirin:
No fixed.
Version 1.18.0-rc.5.0-gb21ab5ad5-wasm32-wasm-clang-12.0.0
![image](https://github.com/epam/Indigo/assets/26869421/2581803c-4680-4dde-afce-f1302cb81bb8)

PR #1633, opened 2024-01-29
 #1536 Add/Remove explicit hydrogens feature doesn't work for reactions on canvas (crash happens)

AliaksandrDziarkach:
Add reaction support. Add UT


even1024 (approved):

even1024 (approved):
looks ok

Issue #1826, opened 2024-03-15
Add sorting and pagination to bingo-elastic 


khyurri:
**Background**
Currently, users can only access a limited count of compounds (up to 10,000 by default) from ElasticSearch. We need to incorporate sorting and pagination functionality into the bingo-elastic driver. This will allow users to receive more than 10,000 compounds from the index.

**Solution**
ElasticSearch offers a pagination feature (PIT). Visit https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html to learn more. 
We need to incorporate both the Java and Python functionality to enable transparent use of PIT, thus eliminating the need for user-level management. Therefore, if a user sets a sorting parameter, the bingo-elastic driver will automatically continue fetching documents until the limit is reached. 
Importantly, after implementing this change, the limit option should be optional. This will allow users to download the entire index if they choose to do so.

khyurri:
In progress PR: https://github.com/epam/Indigo/pull/1712

uladkaminski:
Java changes PR: #1827

uladkaminski:
Additional Java changes PR: #1880

uladkaminski:
More ElasticStream methods implemented: [#1946](https://github.com/epam/Indigo/pull/1946)

PR #1827, opened 2024-03-15
#1826 - Pagination and sorting for ElasticStream

uladkaminski:
## Generic request

Implementation is done for the expected behaviour. Four cases were considered:

- The stream has sorting and limit: return all records. the amout limited by the result hits count or limit parameter.
image

- The steam has sorting only: return all records sorted. (even more then 10k)
image

- The steam has limit only: return the amount of records from the limit parameters but not more than 10k records.
image
image

- The stream has no limit and sorting: return not more then 10k records
image

### DIFF
--- a/bingo/bingo-elastic/java/src/main/java/com/epam/indigo/elastic/ElasticStream.java
+++ b/bingo/bingo-elastic/java/src/main/java/com/epam/indigo/elastic/ElasticStream.java
@@ -4,7 +4,12 @@ import com.epam.indigo.BingoElasticException;
 import com.epam.indigo.model.Helpers;
 import com.epam.indigo.model.IndigoRecord;
 import com.epam.indigo.model.NamingConstants;
-import com.epam.indigo.predicate.*;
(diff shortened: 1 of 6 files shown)


uladkaminski:
Several things I would probably need to clarify: 

- Since we don't actually use Comparator, I introduced IndigoComparator structure to hold field and sort order for comparison. 

- The [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html) approach requires distinct value in the search result. That is why I am adding additional _doc sorting in case of any sorting was added. 

- Current sorted method has accumulative effect. So every time the method is invoked additional sorting is added. Maybe it makes sense to reinitiate the sorting every time and introduce some kind of multiple field sorting?

- Except `sorted(Comparator)` Stream API contains default `sorted()` method which also can have some kind of default sorting logic. Do you have an idea what the logic can be?

- Not sure if this is the issue. Currently sorting by name is not allowed due to name is not a keyword. Let me know if this functionality is expected.

khyurri (approved):

PR #1987, opened 2024-06-05
Pull IndigoRecord attribute as the document ID for index updates in bingo-elastic

rvaidya:
Currently, there is no way to update existing records in ElasticSearch when calling ElasticRepository.indexRecords in bingo.  This leads to lots of duplicate documents being inserted.

This uses the IndigoRecord internalID field as the ID field for the document, so that existing documents can be updated.


rvaidya:
Made the attribute used for document id configurable - can either provide your own mapper from IndigoRecord to document id (if you want to pull from custom objects), or use IndigoRecord.internalID

Also provide optional implementation for auto flushing index

PR #2147, opened 2024-07-23
Load pathway reactions from RDF as a simple list of reactions

nupagadi:
* Load RDF
* Save as KET
* Add new interface for ReactionAutoLoader loadReaction


core/indigo-core/reaction/pathway_reaction.h
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+
+#pragma once
+
+#ifdef _WIN32
+#pragma warning(push)
```
AliaksandrDziarkach, line 22: 1) It is not win32 specific - it is MSVC specific
2) where is "#pragma warning(pop)" ?

nupagadi: 1. You want it to be _MSC_VER? Perhaps, you're correct, but we use _WIN32 all over the code for this matter. And I won't change it in other places. Do you want it to be MSC_VER here and WIN32 in other places?
2. Fixed

AlexanderSavelyev (approved):

Issue #2287, opened 2024-08-30
Wrong smiles in HELM


AliaksandrDziarkach:
Molecules saved to HELM generate SMILES with non-existing cycles:
`PEPTIDE1{A}|CHEM1{[C(N%91)=C%92.[*:1]%92.[*:2]%91 |$;;;_R1;_R2$|}$CHEM1,PEPTIDE1,1:R2-1:R1$$$V2.0`
Also all molecules repited for each polymer.

AlexeyGirin:
Verified.

**Versions**
- Ketcher Version 2.25.0-rc.1 Build at 2024-09-03; 15:31:03
- Indigo Toolkit Version 1.24.0-rc.1.0-gb8f099851-wasm32-wasm-clang-19.0.0
- Chrome Version 128.0.6613.114 (Official Build) (64-bit)
- Win10

PR #2288, opened 2024-08-30
 #2287 Wrong smiles in HELM

AliaksandrDziarkach:
Fix code. Update UT.

## Generic request


AliaksandrDziarkach:
1) Molecule processing was inside `for(auto sequence: sequnces)` block which cause molecule repetion for each sequence.

2) Smiles save by default use `separate_rsites` option which lead to generate fake cycles at R-site atoms. Have to set this option to false to avoid such cycles. btw. when save smiles from ketcher this option enabled by default which lead to `[*:1]C%91.[*:2]%91 |$_R1;;_R2$|` instead of `[*:1]C[*:2] |$_R1;;_R2$|`

AlexanderSavelyev (approved):

PR #2570, opened 2024-10-21
#2407 - The ordering of branches of cascade reactions and reactions themselves should be the same on Load/Save from/to RDF

even1024:
## Generic request
- [x] PR name follows the pattern `#1234 – issue name`
- [x] branch name does not contain '#'
- [x] base branch (master or release/xx) is correct
- [x] PR is linked with the issue
- [x] task status changed to "Code review"
- [x] code follows product standards
- [x] regression tests updated

Issue #2407, opened 2024-09-19
The ordering of branches of cascade reactions and reactions themselves should be the same on Load/Save from/to RDF 


ivanmaslow:
**Expected pathway reaction ordering**
Idea is to follow reactant’s order: first reactant in a reaction must be "at the top" position of a multi-tailed arrow, as upper molecule in a reaction (reactant near multi-tailed arrow) must be the first reactant in RXN format.

![Image](https://github.com/user-attachments/assets/bf5cb4d7-1e11-4231-a7a8-e5ff2922393f)

**Steps to Reproduce 1**
1. Load from RDF file with the single cascade reaction with several branches:
[!Synthia_path_5_tails.zip](https://github.com/user-attachments/files/17058376/Synthia_path_5_tails.zip)

**Actual Result**
- The first branch of cascade reactions in RDF file is the second on Canvas:
![image](https://github.com/user-attachments/assets/052b516b-7ed2-4e59-9ce5-038d9a41ad40)

**Expected Result**
- The first branch of cascade reactions in the RDF file is the first on Canvas

2. Save to RDF and Load again

**Actual Result**
- Positioning of elements is changed after Save/Load, but it should like after the first load:
![image](https://github.com/user-attachments/assets/5862ef38-30b3-405a-b2e7-4068727dccb9)

**Expected Result**
- Positioning of elements isn't changed after Save/Load, it's displayed like after the first load

**Steps to Reproduce 2**
1. Load from RDF file with several single reactions:
[rdf-rxn-v2000-reactions-1x1-2x2-3x3.zip](https://github.com/user-attachments/files/17058595/rdf-rxn-v2000-reactions-1x1-2x2-3x3.zip)

**Actual Result**
- The first reaction in RDF file is the third on Canvas, the third reaction in RDF is the first on Canvas:
![image](https://github.com/user-attachments/assets/4721e11a-d4a4-4683-8258-994e22358b2b)

**Expected Result**
- Reactions should be displayed on Canvas as in RDF file 1:1, 2:2, 3:3

2. Save to RDF

**Actual Result**
- It can't be saved as separate reactions:
![image](https://github.com/user-attachments/assets/59ef829e-6262-479b-ac09-538fc2297c00)

**Expected Result**
- It can be saved as separate reactions 1:1, 2:2, 3:3
- It will be fixed in scope of https://github.com/epam/Indigo/issues/2412

**Steps to Reproduce 3 - Ordering of elements**
[...]

PR #2577, opened 2024-10-24
Backmerge: #2407 - The ordering of branches of cascade reactions and reactions themselves should be the same on Load/Save from/to RDF (#2570)

even1024:
## Backmerge request
- [x] PR name follows the pattern `Backmerge: #1234 – issue name`
- [x] code contains only backmerge changes


AliaksandrDziarkach (approved):
LGTM

PR #2633, opened 2024-11-14
bugfix: indentation errors in docstring description.

netluo:
bugfix: indentation errors in docstring description.

When accessing to Swagger(http://host:port/apidocs/), it throws an exception - "Failed to load API definition" .
Due to indentation errors in the two files.


netluo:
will open a new pr.

Issue #2739, opened 2025-01-24
Wrong stereochemistry when loading cdxml reaction


jblack-mestre:
**Summary**
When creating an .svg file with indigo-depict and wrong_stereochemistry.cdr, two identical product molecules are shown in the reaction, instead of two different isomers. The second product should have an upwards wedge, not a downwards dash.

**Steps to Reproduce**
1. indigo-depict.exe wrong_stereochemistry.cdr wrong_stereochemistry.svg

**Actual behavior**
Indigo is reusing the stereochemistry data from the first product with the second product.
![Image](https://github.com/user-attachments/assets/74991da2-5b51-4e39-a50a-302caa4d6e1f)

**Expected behavior**
Indigo should clear the stereochemistry data from the first product, so the correct stereochemistry can be used in the second product.
![Image](https://github.com/user-attachments/assets/a78cb3f0-f089-4f23-81c4-d007036eb111)

**Environment details:**
 - Windows 11
 - Indigo, commit 4960b8e

**Attachments**
[wrong_stereochemistry.txt](https://github.com/user-attachments/files/18533909/wrong_stereochemistry.txt)

**Additional context**
In MoleculeStereocenters::clear(), both the _stereocenters and _atropocenters maps need to be cleared

Comment from QA:
Reproduced on Indigo Version 1.28.0-rc.3.0-gd5bcc27cb-wasm32-wasm-clang-19.0.0

AlexeyGirin:
Verified.
![Image](https://github.com/user-attachments/assets/300a394e-8a2c-4dbb-80f3-6fa84e609243)

- Ketcher Version 3.1.0-rc.1 Build at 2025-02-04; 21:27:02
- Indigo Version 1.29.0-rc.1.0-g2e30df8f7-wasm32-wasm-clang-19.0.0
- Chrome Version 133.0.6943.54 (Official Build) (64-bit)
- Win10

PR #2740, opened 2025-01-24
#2739 - Wrong stereochemistry when reading cdxml reaction

jblack-mestre:
## Generic request

### DIFF
--- a/core/indigo-core/molecule/molecule_cdxml_loader.h
+++ b/core/indigo-core/molecule/molecule_cdxml_loader.h
@@ -189,6 +189,7 @@ namespace indigo
     class BaseCDXProperty
     {
     public:
+        virtual ~BaseCDXProperty() = default;
         virtual bool hasContent() const = 0;
         virtual std::unique_ptr<BaseCDXProperty> copy() = 0;
         virtual std::unique_ptr<BaseCDXProperty> next() = 0;
@@ -375,6 +376,7 @@ namespace indigo
     class BaseCDXElement
     {
     public:
+        virtual ~BaseCDXElement() = default;
         virtual bool hasContent() = 0;
         virtual std::unique_ptr<BaseCDXElement> copy() = 0;
         virtual std::unique_ptr<BaseCDXProperty> firstProperty() = 0;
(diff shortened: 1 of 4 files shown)


AlexanderSavelyev:
could you please create at least one test (unit or regression) so that any new version does not break the change in the future
test can be added into unit tests here
https://github.com/epam/Indigo/blob/master/core/indigo-core/tests/tests/formats.cpp
or regression test. e.g.
https://github.com/epam/Indigo/blob/master/api/tests/integration/tests/formats/cdx.py

jblack-mestre:
I figured out how to get the unit test working - let me know if this is ok or I need to change it

AlexanderSavelyev:
thanks for the updated PR, but I think comparing two cdxml is not very stable. If Indigo changes any other piece of model - it will lead to different result. I would rather just calculate number of enhanced stereo
using stereocenters iterator for molecules .eg. this one
https://github.com/epam/Indigo/blob/master/core/indigo-core/molecule/src/molfile_saver.cpp#L802
or if it is hard at least save to extendend SMILES
https://github.com/epam/Indigo/blob/master/core/indigo-core/tests/tests/formats.cpp#L157

jblack-mestre:
In this case we need to count the different bond directions in the products. Both products have 1 OR stereocenter, but the first isomer should have a dashed wedge and the second should have a solid wedge at this stereocenter. Before the fix, both products had dashed wedges. Let me know if you think this is ok - if not, I can try the extended SMILES.

AlexanderSavelyev:
yes, this looks much better. I started the CI and after successful run will merge

AlexanderSavelyev:
yes, we see the issue. We will create a patch for you and just need to merge it into your branch

AlexanderSavelyev:
Hi @jblack-mestre , could you please merge changes from this branch "fix_cdx_desctructors". After that the warning should be fixed

jblack-mestre:
thanks, done

AlexanderSavelyev (approved):

AlexanderSavelyev:
all tests passed. I merged the PR. The change will be available in the next version (approx release date is next month)
@jblack-mestre  Thanks so much for the fixes!

Issue #2741, opened 2025-01-24
Image missing stereochemistry information when using abbreviations


jblack-mestre:
**Summary**
When generating an .svg image with indigo-depict missing_stereochemistry_info.cdr, the product is missing a downward dash bond connecting to the CN. This problem doesn't happen when the CN is displayed in expanded notation.

**Steps to Reproduce**
1. indigo-depict.exe missing_stereochemistry_info.cdr missing_stereochemistry_info.svg

**Actual behavior**
The svg image is missing a downward dash bond in the product when connecting to the CN group.
![Image](https://github.com/user-attachments/assets/1fddbd59-3b34-4c39-96c2-009dbfd58aed)

**Expected behavior**
Indigo should display the correct steroechemistry as described in the cdxml file.
![Image](https://github.com/user-attachments/assets/b905f618-d25d-4e02-ac8e-46db33fbac97)

**Environment details:**
 - windows 11
 - indigo [4960b8e]

**Attachments**
[missing_stereochemistry_info.txt](https://github.com/user-attachments/files/18534877/missing_stereochemistry_info.txt)
[expanded_not_missing_steroechemistry.txt](https://github.com/user-attachments/files/18534878/expanded_not_missing_steroechemistry.txt)

**Additional context**
In  MoleculeRenderInternal::_prepareSGroups, the bond direction should be set when creating a new bond between the super atom and the neighboring atom.

Comment from QA:
Reproduced on Indigo Version 1.28.0-rc.3.0-gd5bcc27cb-wasm32-wasm-clang-19.0.0

AlexeyGirin:
Verified.
![Image](https://github.com/user-attachments/assets/ba852a57-b587-4e72-9dd0-ab3dff04418c)

- Ketcher Version 3.2.0-rc.1 Build at 2025-03-04; 14:24:35
- Indigo Version 1.30.0-rc.1.0-g9bdedaba9-wasm32-wasm-clang-19.0.0
- Chrome Version 134.0.6998.36 (Official Build) (64-bit)
- Win10

AlexeyGirin:
@jblack-mestre - issue has been fixed and released soon (~ in month)

jblack-mestre:
Thanks

PR #2742, opened 2025-01-24
#2741 Image missing stereochemistry information when using abbreviations

jblack-mestre:
## Generic request

### DIFF
--- a/core/render2d/src/render_internal.cpp
+++ b/core/render2d/src/render_internal.cpp
@@ -874,8 +874,9 @@ void MoleculeRenderInternal::_prepareSGroups(bool collapseAtLeastOneSuperatom)
                                     else
                                     {
                                         Molecule& amol = mol.asMolecule();
-                                        bondID = amol.addBond(superAtomID, neighboringAtomID, amol.getBondOrder(neighboringBondID));
+                                        bondID = amol.addBond(neighboringAtomID, superAtomID, amol.getBondOrder(neighboringBondID));
                                         amol.setEdgeTopology(bondID, amol.getBondTopology(neighboringBondID));
+                                        amol.setBondDirection(bondID, mol.getBondDirection(neighboringBondID));
                                     }
                                     if (_bondMappingInv.find(bondID) != _bondMappingInv.end())
                                         _bondMappingInv.erase(bondID);


jblack-mestre:
Is is possible to write a unit test in api/cpp/tests/rendering/basic.cpp for this bug and others? These changes affect render_internal.cpp

AlexanderSavelyev:
I guess in this particular case it is not render issue. It is just missing stereocenter, which is worth to test using just stereo information (e.g. number of stereo bonds down)

jblack-mestre:
Unfortunately we can't count the up and down bonds for the unit test; in this case, indigo correctly assigns the stereocenters with the correct bond directions, but when the molecule is passed through MoleculeRenderInternal::_prepareSGroups(), the stereocenter to the super atom is forgotten.
Can you suggest another place for a unit test?

AlexanderSavelyev:
for render tests we use regression test with pixel comparison
e.g. https://github.com/epam/Indigo/blob/master/api/tests/integration/tests/rendering/render_reactions.py#L108
in order to execute regression tests need to build indigo-python target
```
cmake --build . --config Release --target indigo-python
# copy
python -m pip install ..\dist\epam.indigo-1.29.0.dev1-py3-none-win_amd64.whl
# run tests by mask
python3 ..\api\tests\integration\test.py -p render_reactions
```

also, there are some requirements should be installed to python 
```
python3 -m pip install Pillow
```

AlexanderSavelyev:
please also merge latest changes from `master` branch

jblack-mestre:
thanks for the info, I've added a test to render_reactions.py.

I merge the changes from master yesterday. Git merge upstream/master is telling me I am up to date

jblack-mestre:
Do you know what the problem is? My fork is in sync with indigo/master and my branch is up to date with my fork's master and the indigo/master branches

AlexanderSavelyev:
I just verified, your branch is not up-to-date
please do 
git fetch upstream
git merge upstream/master

jblack-mestre:
git_fetch

AlexanderSavelyev:
oh, I see, Indigo team didn't backmerge to master the changes about github actions. Will fix it and let you know

AlexanderSavelyev:
please take the latest changes from master

jblack-mestre:
I don't have access to a linx/mac machine in order to generate the reference files for the regression test. How is this normally done?

The windows tests fail due to a diff in the output - do I need to update a output reference file?

AlexanderSavelyev:
* no need to merge from master anymore. It was just one issue with checkout github action upgrade. There is no need to PR to be up-to-date with master branch (and in general it is not possible)
* for windows it seems that the test is running ok, but the console output is different. please replace the expected output from "out" folder into "api/tests/integration/ref/rendering/render_reactions.py.out"
* you can check if test is passed by running python3 api\tests\integration\test.py -p render_reactions locally
* the image from win folder should be copied into linux and mac. Usually the images are the same, but if there are difference - the test will print out the expected result

jblack-mestre:
thanks for the info - I've made the changes

AlexanderSavelyev:
all tests passed, merged, thanks so much for the fix!

PR #2745, opened 2025-01-29
Add virtual destructors into CDX to avoid memory leak

AliaksandrDziarkach:
To correct destruct derived classes base class destructor should be virtual,


AlexanderSavelyev:
closed via #2740

Issue #425, opened 2021-06-13
Smiles with attachment points is not read correctly (valences are wrong)


achher:
When atachment points is located on a non carbon atom then valence of those atoms is sometimes wrong when reading canonical smiles (generated by Indigo). See sceenshot attached:
Column "Structure" is the original molfile. Column "Canonical Smiles" is the generated Canonical Smiles by Indigo. Canonical Smiles is correct. When Indigo reads the Canonical Smiles then structure in column "Structure (Canonical Smiles)" is displayed. Which is not the original structure anymore. When Indigo generates Canonical Smiles from wrong structure, then Smiles in column "Canonical Smiles (After)" is generated. Column "Canonical Smiles" and column "Canonical Smiles (After)" are oviously not the same.
![grafik](https://user-images.githubusercontent.com/70947872/121803994-5b81ae80-cc44-11eb-9c77-05e142904746.png)

Canonical Smiles used for this example are (column "Canonical Smiles"):
```
[O-]%91.[*:1]%91 |$;_AP1$|
[S-]%91.[*:1]%91 |$;_AP1$|
S%91C(C1C=CC=CC=1)C1C=CC=CC=1.[*:1]%91 |$;;;;;;;;;;;;;;_AP1$|
[SeH]%91.[*:1]%91 |$;_AP1$|
[Se]%91C1C=CC=CC=1.[*:1]%91 |$;;;;;;;_AP1$|
[SiH3]%91.[*:1]%91 |$;_AP1$|
[NH3+]%91.[*:1]%91 |$;_AP1$|

```
Canonical Smiles generated after reading the original Canonical Smiles is read for this example are (column "Canonical Smiles (After)"):
```
[OH2-]%91.[*:1]%91 |$;_AP1$|
[SH-]%91.[*:1]%91 |^1:0,$;_AP1$|
[SH2]%91C(C1C=CC=CC=1)C1C=CC=CC=1.[*:1]%91 |$;;;;;;;;;;;;;;_AP1$|
[SeH2]%91.[*:1]%91 |^1:0,$;_AP1$|
[SeH]%91C1C=CC=CC=1.[*:1]%91 |^1:0,$;;;;;;;_AP1$|
[SiH4]%91.[*:1]%91 |$;_AP1$|
[NH4+]%91.[*:1]%91 |$;_AP1$|
```

AlexeyGirin:
Verified.
| Structure | Canonical Smiles | Canonical Smiles (after) | Structure (after) | Comment |
[...]

Issue #2746, opened 2025-01-30
Incorrect phosphorus valence


jblack-mestre:
**Summary**
When generating an .svg image with indigo-depict phosphorus_valence.cml, the phosphorus is incorrectly depicted as having a wrong valence of 5. It should have a valence of 3, with no added hydrogens.

**Steps to Reproduce**
1. indigo-depict.exe phosphorus_valence.cml phosphorus_valence.svg

**Actual behavior**
The phosphorus is depicted as having a valence of 5 with two added hydrogens.
![Image](https://github.com/user-attachments/assets/f989381c-8aa0-4b2d-bc0f-572875ff7091)

**Expected behavior**
The phosphorus should be depicted as having a valence of 3 with no added hydrogens.
![Image](https://github.com/user-attachments/assets/6a18e0fd-6c0d-44c3-8dc7-c9065fea552d)

**Environment details:**
 - windows 11
 - indigo [4960b8e](https://github.com/epam/Indigo/commit/4960b8efdc1d1d394b362aee65e2ce931aa03f4e)

**Attachments**
[phosphorus_valence.cml.txt](https://github.com/user-attachments/files/18600572/phosphorus_valence.cml.txt)

**Additional context**
When adding sgroup bonds, indigo recalculates the valence and implicit hydrogens before it has removed the bond to the atom connecting to the sgroup. So P initially has a valence of 3, then this is raised to 4 when adding the bond to the new sgroup. Once the valence is recalculated, this is increased to 5.

AliaksandrDziarkach:
Looks like issue reproduced when abbreviation contracted and all valence connection used:

![Image](https://github.com/user-attachments/assets/b769677c-f8a0-4e99-a0b0-a27e92e901c6)

![Image](https://github.com/user-attachments/assets/b4496946-68d7-4e32-b540-f63764bcbd33)

AliaksandrDziarkach:
Just add correspondig count of `C`, then make one of them `Superatom S-Group` and contract it.

```
{
    "root": {
        "nodes": [
            {
                "$ref": "mol0"
            },
            {
                "$ref": "mol1"
            }
        ],
        "connections": [],
        "templates": []
    },
    "mol0": {
        "type": "molecule",
        "atoms": [
            {
                "label": "P",
                "location": [
                    9.850000000000001,
                    -4.575,
                    0
                ]
            },
            {
                "label": "C",
                "location": [
                    10.850000000000001,
                    -4.575,
                    0
                ]
            },
[...]

PR #2783, opened 2025-02-24
#425 Smiles with attachment points is not read correctly (valences are wrong)

jblack-mestre:
## Generic request


jblack-mestre:
Similar bug to https://github.com/epam/Indigo/issues/2746; need to remove the old bonds before removing atoms (and thereby recalculating the implicit hydrogens)

core/indigo-core/molecule/src/smiles_loader.cpp
```diff
@@ -534,9 +534,17 @@ void SmilesLoader::_readOtherStuff()
...
                         for (k = v.neiBegin(); k != v.neiEnd(); k = v.neiNext(k))
+                        {
                             _bmol->addAttachmentPoint(rnum, v.neiVertex(k));
+                            if (_bmol->findEdgeIndex(i, v.neiVertex(k)) >= 0)
+                            {
+                                bondsToRemove.push(_bmol->findEdgeIndex(i, v.neiVertex(k)));
+                            }
+                        }
                         to_remove.push(i);
+                        _bmol->removeBonds(bondsToRemove);
```
AliaksandrDziarkach, line 547: Maybe better fix `void BaseMolecule::removeAtoms(const Array<int>& indices)` to remove bonds connected to removed atoms?
Because edges without vertex looks weird.

jblack-mestre:
I don't understand why the tests fail with:
{F39E453B-DD69-4CAE-8474-BC9CA3CE4D50}

AliaksandrDziarkach:
> I don't understand why the tests fail with: 
It is unclear without context, but I already hit same issue - it is due to different key order in iron python, so line moved to different position.
I added comment to code.

AlexanderSavelyev (approved):

PR #2784, opened 2025-02-24
Atomic level feature draft

StLeonidas:
## Remove-me-sectionAtomic level feature draft
* Recovers atomic level structure from ket file
- [x] exposed to wasm
- [x] draft works with MolV3000
- [x] draft works with MolV3000linear peptides


core/indigo-core/molecule/src/atomic_level.cpp
```diff
@@ -0,0 +1,306 @@
...
+float findAngle(float dx, float dy) {
+    float angle;
+    if (dx == 0)
+      angle = dy > 0 ? 0 : PI;
+    else if (dy == 0)
+      angle = dx > 0 ? -PI / 2 : PI / 2;
+    else {
+      float tan = dy / dx;
+      float at = atan(tan);
+      angle = (dx < 0) ? PI / 2 + at : -PI / 2 + at;
```
drizhina, line 58: might be good idea to specify that function finds angle between point and Y axis. also, I believe atan2 is prefered

StLeonidas: Yes, thanks. Possibly will refine all rotation to rays not setions as well

AliaksandrDziarkach: I think you could use `float Vec2f::tiltAngle()`

core/indigo-core/molecule/src/atomic_level.cpp
```diff
@@ -0,0 +1,306 @@
+#include "molecule/atomic_level.h"
+using monomers_map = std::map<std::string, std::unique_ptr<KetBaseMonomer>>;
+using attach_map = std::map<std::string, KetAttachmentPoint>;
+
+std::string strReplace(std::string str, const std::string& substr, const std::string& tostr, const bool first = false) {
+    if (substr.size() > str.size())
+      return str;
+  
+    size_t detectionDomain = str.size() - substr.size() + 1;
+    detectionDomain = (substr == "\\t" || substr == "\\r" || substr == "\n") ? detectionDomain + 1 : detectionDomain;
```
skalkin, line 10: Are you sure about \\n and \\r, but \n (only one "\")?

core/indigo-core/molecule/src/atomic_level.cpp
```diff
@@ -0,0 +1,306 @@
...
+  
+    bool stop = false;
+  
+    if (substr.size() == 0 || detectionDomain <= 0)
+      return str;
+  
+    std::string res;
+    for (size_t i = 0; i < str.size(); ++i) {
+      if (stop) {
+        res += str[i];
```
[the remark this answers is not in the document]
StLeonidas: I will refine to classical str replace as additional feature of replace first is not required any more

AliaksandrDziarkach:
> I wonder the reasons KetAtom is inherited class while monomer libraries contain KetAtomBase class, are there any cases when atoms in the library have no sufficient data to be KetAtom?
It also can be `Atom List` or `RG label`
image

Issue #2150, opened 2024-07-24
Can't save a schema with some elements from Periodic Table in PNG and SVG format


AnnaKosol:
**Steps to Reproduce**
1. open Ketcher
2. Add Benzene ring
3. Add RG (Roentgenium, 111 from Periodic Table)
4. Save in PNG or SVG format

**Expected behavior**
<img width="148" alt="Снимок экрана 2024-07-24 в 15 59 39" src="https://github.com/user-attachments/assets/8a333704-5a80-41ce-8b53-d3507d7a350e">

**Actual behavior**
An error occurred.
<img width="446" alt="Снимок экрана 2024-07-24 в 16 00 20" src="https://github.com/user-attachments/assets/d01d75b6-2457-4aa6-a5a4-83f8c025df2d">

**The same situation is for:**
Mt (Meitnerium, 109 from Periodic Table)
<img width="670" alt="Снимок экрана 2024-07-24 в 16 28 36" src="https://github.com/user-attachments/assets/3b3e5012-fdaf-4501-91aa-df624f2e4cbc">

**Environment details:**
Ketcher Version 2.23.0-rc.3 Build at 2024-07-14; 20:34:10
Indigo Version 1.22.0-rc.3.0-gbb5c51fc2-wasm32-wasm-clang-19.0.0
Chrome Версия 126.0.6478.127 (Официальная сборка), (arm64)
MacOs Sonoma 14.5

Zhirnoff:
Tested. Bug fixed. But colors not match to Periodic table.

![Image](https://github.com/user-attachments/assets/127a0499-caec-4e37-9b84-41d40ebb2797)
![Image](https://github.com/user-attachments/assets/5618a992-6385-408a-8d03-b4ea1b39d8f4)

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 133.0.6943.127 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.3.0-rc.1]
**Indigo version** 
[Version 1.31.0-rc.1]

PR #2786, opened 2025-02-25
 #2150 Can't save a schema with some elements from Periodic Table in PNG and SVG format

jblack-mestre:
## Generic request


jblack-mestre:
I just extended the colours so the heavy atoms will render to png. Let me know if I should give these a specific colour, or if black if ok. Also let me know if I should add test.

AlexanderSavelyev (approved):

Issue #2900, opened 2025-05-12
CIP labels are always rendered from cdxml


jblack-mestre:
**Summary**
CIP labels are always rendered from cdxml, even when they are set not to display

**Steps to Reproduce**
1. indigo-depict.exe cip_labels.cdx cip_labels.svg

**Actual behavior**
All CIP labels are displayed, when only the labels on the first molecule have been explicitly set to be displayed.
![Image](https://github.com/user-attachments/assets/d1666ff0-74ec-4637-a997-80e6ad480b8f)

**Expected behavior**
The CIP labels on the first molecule should be displayed, while the labels of the second molecule should be hidden.
![Image](https://github.com/user-attachments/assets/ea8b04c1-f751-4c67-9cd1-7a25d8127d09)

**Environment details:**
 - windows 11
 - indigo [098fd1d](https://github.com/epam/Indigo/commit/098fd1d3567a82343d89e3fab4c19f35227d0cc7)

**Attachments**
[cip_labels.cdx.txt](https://github.com/user-attachments/files/20158003/cip_labels.cdx.txt)

**Additional context**
We need to parse the ShowAtomStereo="yes" from the cdxml file. The default for this option should be "no".

mobilisf:
### **1. Problem Description:**  
The CIP (Cahn-Ingold-Prelog) labels are always rendered from a CDXML file, even when the CDXML explicitly specifies that these labels should not be displayed. This happens because the `ShowAtomStereo="no"` attribute in the CDXML file is ignored during rendering. As a result, all CIP labels (e.g., `(R)`, `(S)`) are displayed, regardless of whether they are set to be shown or hidden.

#### Hypothesis:  
The rendering logic does not correctly parse the `ShowAtomStereo` attribute within the CDXML file. Instead, it defaults to displaying all CIP labels unless specifically overridden. This likely occurs due to the absence of logic to process and enforce `ShowAtomStereo="no"` during parsing.

---

### **2. Steps to Reproduce:**  
1. Use the `indigo-depict.exe` command-line tool to generate an SVG file from a CDXML file:  
   ```
   indigo-depict.exe cip_labels.cdx cip_labels.svg
   ```

2. Open the generated `cip_labels.svg` file in any SVG viewer or browser.  
3. Observe the CIP labels rendered on both molecules.  

   - The first molecule contains CIP labels explicitly set to display (`ShowAtomStereo="yes"`).  
[...]

PR #2907, opened 2025-05-13
#2900 CIP labels are always rendered from cdxml

jblack-mestre:
## Generic request


jblack-mestre:
CIP labels by default are no longer displayed unless the showAtomStereo="yes" attribute is present in the cdxml tag. This change will only affect rendering from cdxml files.

jblack-mestre:
Hi @AlexanderSavelyev, the problem is that the user can choose to turn off and on specific cip labels, so if we want to render a cdxml file we have to note which atoms have a cip label visible and which don't. For example: 
{31C99CAD-CD3F-4DCD-9D48-EBDE0862929E}
It is a strange example; normally you want to display all the cip labels or you don't. Unfortunately, we have to know which cip labels are visible on which atoms.

What do you think? If you think we shouldn't have the std::unordered_map, I can replace it will a single boolean.

AliaksandrDziarkach (approved):

PR #2950, opened 2025-05-26
Monomers expand

StLeonidas:
Full expand support:
[x] cases with etended and not extended monomers in a single entity
[x] linear graphs propogation
[x] cycled structures (>1 cycles)
[x] branched structures

Bottlenecks:
visual inspenction and polishing required
cpp part api is not standard (in wasm - indogo-ketcher.cpp)
up to 3 R groups supported (the most cases, but htere can be more)
cycles that are not round could be modified


AliaksandrDziarkach:
Unit tests added.
Changes merged to master as part of https://github.com/epam/Indigo/pull/3021

Issue #3206, opened 2025-09-24
Export (or import) to (from) Mol v3000 works wrong for CHEM monomers


AlexeyGirin:
**Steps to Reproduce**
1. Go to Macro - Flex mode (clean canvas)
2. Load using Paste from clipboard way: `CHEM1{[5TAMRA]}$$$$V2.0`
<img width="134" height="136" alt="Image" src="https://github.com/user-attachments/assets/8147d1e7-2b75-4e3d-80ae-1f19c1a66875" />

3. Export canvas to `Mol v3000`
4. Load export result back to canvas
<img width="150" height="127" alt="Image" src="https://github.com/user-attachments/assets/5df3f260-83d7-421d-8884-ce625eea86c3" />

5. Save canvas to HELM

**Actual behavior**
Export result: `CHEM1{[C1(=[N+](/C)\C)\C=CC2=C(C3=CC(C(=O)[*:1])=CC=C3C(=O)[O-])C3=CC=C(N(C)C)C=C3OC2=C\1 |$;;;;;;;;;;;;;_R1;;;;;;;;;;;;;;;;;;$|]}$$$$V2.0`
<img width="469" height="438" alt="Image" src="https://github.com/user-attachments/assets/ddbea74c-426e-4865-818a-b00d115126c0" />

**Expected behavior**
Export result: `CHEM1{[5TAMRA]}$$$$V2.0`
<img width="475" height="433" alt="Image" src="https://github.com/user-attachments/assets/86792e3c-b109-4242-9fb1-913021f7984f" />

This probably happens because system monomer type from `CHEM` to `Linker` after save to mol (or after load from mol)

<img width="493" height="344" alt="Image" src="https://github.com/user-attachments/assets/8a9c7a92-4533-4808-811a-cc54fac0c911" />

**Versions**
- Ketcher Version 2.27.0-rc.1 Build at 2024-11-05; 14:55:18
- Indigo Toolkit Version 1.26.0-rc.1.0-g904d2d992-wasm32-wasm-clang-19.0.0
- Chrome Version 130.0.6723.117 (Official Build) (64-bit)
- Win10

Related issue: https://github.com/epam/Indigo/issues/1381

AlexeyGirin:
Desided to load Linker and Terminator as CHEM

AlexeyGirin:
Verified.
[...]

PR #3207, opened 2025-09-24
 #3206 - Export (or import) to (from) Mol v3000 works wrong for CHEM monomers

AliaksandrDziarkach:
Convert linker to chem at mol file load. Fix UT

## Generic request


core/indigo-core/molecule/src/molfile_loader.cpp
```diff
@@ -2907,7 +2907,10 @@ void MolfileLoader::_readCtab3000()
                     QS_DEF(Array<char>, temp_class);
                     strscan.readWord(temp_class, 0);
                     temp_class.push(0);
-                    _bmol->setTemplateAtomClass(i, temp_class.ptr());
```
AlexanderSavelyev, line 2910: change in ket saver, not in molfile loader

AliaksandrDziarkach: Fixed

Issue #3187, opened 2025-09-17
Export to HELM works wrong for custom monomers imported from HELM with inline SMILES (part 2)


AlexeyGirin:
**Steps to Reproduce**
1.Switch to Macromolecules mode - Flex canvas
2. Load from HELM: `RNA1{[O1[C@@H]%91[C@H](O)[C@H](O%92)[C@H]1CO%93.[*:3]%91.[*:1]%93.[*:2]%92 |$;;;;;;;;;_R3;_R1;_R2$|]p}$$$$V2.0`
<img width="507" height="484" alt="Image" src="https://github.com/user-attachments/assets/3aff7fd8-5a73-4211-b347-2023efe1e54b" />

3. Save canvas to HELM

**Actual behavior**
Export result: `RNA1{[O1[C@H](CO[*:1])[C@@H](O[*:2])[C@@H](O)[C@H]1[*:3] |$;;;;_R1;;;_R2;;;;_R3$|].p}$$$$V2.0`
<img width="448" height="421" alt="Image" src="https://github.com/user-attachments/assets/d23ef93e-3b29-42d2-a192-e2733dad0357" />

**Expected behavior**
Export result: `RNA1{[O1[C@H](CO[*:1])[C@@H](O[*:2])[C@@H](O)[C@@H]1[*:3] |$;;;;_R1;;;_R2;;;;_R3$|].p}$$$$V2.0`
<img width="455" height="425" alt="Image" src="https://github.com/user-attachments/assets/cc0504c1-b1c4-4232-b097-895e61f4c5c6" />

**Environment details:**
- Ketcher Version 3.8.0-dev.1 Build at 2025-09-01; 15:10:48
- Indigo Version 1.37.0-dev.1.0-ged340212-wasm32-wasm-clang-19.0.0
- Chrome Version 139.0.7258.155 (Official Build) (64-bit)
- Win10

Related issue: https://github.com/epam/Indigo/issues/2189

AlexeyGirin:
Verified.
HELM: `RNA1{[O1[C@H](CO[*:1])[C@@H](O[*:2])[C@@H](O)[C@@H]1[*:3] |$;;;;_R1;;;_R2;;;;_R3$|].p}$$$$V2.0`
<img width="417" height="463" alt="Image" src="https://github.com/user-attachments/assets/403a0cdb-6dc9-4545-8d19-3478e600ac4f" />

[...]

PR #3214, opened 2025-09-30
 #3187 - Export to HELM works wrong for custom monomers imported from HELM with inline SMILES (part 2)

AliaksandrDziarkach:
Fix code. Add UT

## Generic request


core/indigo-core/molecule/src/sequence_saver.cpp
```diff
@@ -898,7 +899,9 @@ void SequenceSaver::add_monomer(KetDocument& document, const std::unique_ptr<Ket
             }
             sgroups.remove(i);
         }
-        std::string smiles;
+        StandardizeOptions std_opts;
```
AlexanderSavelyev, line 902: should not use standardize to resolve the issue

AlexanderSavelyev (changes requested):

PR #3530, opened 2026-03-05
Add PostgreSQL 18

AliaksandrDziarkach:
Add PostgreSQL 18

## Generic request


Kuppit:
Hi @AliaksandrDziarkach, we build our own PostgreSQL + Bingo image and wanted to move it to 18, so we tried this branch's changes and hit a hard crash.

The two TupleDescAttr changes are necessary but not sufficient. In formIndexTuple() (bingo_pg_buffer.cpp) the attribute is patched by hand, and since 18 the values index_form_tuple() actually reads live in the parallel CompactAttribute array, so attlen stays 0 and the tuple gets formed with a bogus size. Any CREATE INDEX ... USING bingo_idx then kills the backend with "malloc(): invalid size (unsorted)" and SIGABRT. Postgres expects populate_compact_attribute() to be called after a direct write to the FormData_pg_attribute, and there is a comment about exactly this mistake above verify_compact_attribute() in src/backend/access/common/tupdesc.c.

One line fixes it:

```c
#if PG_VERSION_NUM / 100 >= 1800
        TupleDescAttr(index_desc, 0)->attlen = size;
        TupleDescAttr(index_desc, 0)->attalign = 'c';
        TupleDescAttr(index_desc, 0)->attbyval = false;
        populate_compact_attribute(index_desc, 0);
#elif PG_VERSION_NUM / 100 >= 1100
```

We took the three changes from this branch (the two accessors plus sql/18) on top of indigo-1.45.0, built on AlmaLinux 8 against postgresql18-devel 18.6 and ran it on PostgreSQL 18.6, with pytest --db postgres in bingo/tests:
- without the extra call: 31113 passed, 3039 errors, which lines up with the 31113 / 3036 of the last CI run here
- with it: 34152 passed, 0 failed, same count as 14 to 17

I opened #3852 with just that one line on top of this branch.

Auto-EPM-LSOP-Support (approved):

Issue #2109, opened 2024-07-09
Two V3000 export issues: implict H cause wrong index on s groups, and reversed order for COLLECTON / SGroup Blocks.


sapiosciences-dev:
**Summary**
Inside molfile_saver.cpp, function: void MolfileSaver::_writeCtab(Output& output, BaseMolecule& mol, bool query):

1. Issue in V3000 implicit h toggle causing the the S group total count referenced in memory to be modified, causing a failure in parsing the S group when reading from RDKit as it thinks the file is corrupted.

2. Issue in V3000 export of M  V30 BEGIN COLLECTION block occuring before M  V30 BEGIN SGROUP block, causing RDKit failing to parse the section.

**Steps to Reproduce**
Import the RxN . Export and import to RDKit. See Failures. The test file had been provided under attachment section.

**Expected behavior/Actual Behavior**
### Issue 1
First on the issue of improper S group counting:
```
else if (add_implicit_h && Molecule::shouldWriteHCount(mol.asMolecule(), i) && hcount > 0)
        {
            int sg_idx;

            sg_idx = mol.sgroups.addSGroup(SGroup::SG_TYPE_DAT);
            DataSGroup& sgroup = (DataSGroup&)mol.sgroups.getSGroup(sg_idx);

            sgroup.atoms.push(i);

            QS_DEF(Array<char>, tmp_buf);
            ArrayOutput tmp_out(tmp_buf);
            tmp_out.printf("IMPL_H%d", hcount);
            tmp_buf.push(0);
            sgroup.data.readString(tmp_buf.ptr(), true);

            sgroup.name.readString("MRV_IMPLICIT_H", true);
            sgroup.detached = true;
        }
```
The section of the code above adds new s groups to mol.sgroups object while the execution is still in progress exporting the data, and more importantly, this was added **AFTER** the following line was already exported to the text file I/O:

```
    output.writeStringCR("M  V30 BEGIN CTAB");
    output.printfCR("M  V30 COUNTS %d %d %d 0 0", mol.vertexCount(), mol.edgeCount(), mol.countSGroups());
    output.writeStringCR("M  V30 BEGIN ATOM");
```

So, when implicit H is added, total S group counter written already in an earlier line had become incorrect.

### Issue 2

Per this technical manual:
https://www.daylight.com/meetings/mug05/Kappler/ctfile.pdf
[...]

PR #3531, opened 2026-03-06
Fix EPAM Indigo parsing so that the S groups needs to saved before collection block

sapiosciences-dev:
Participants consulted

Matthew Sage, Roles Portfolio Engineering Senior Manager, BIOVIA

Rob Brown, VP, Product Design, Sapio

Yechen Qiao, Core Product Developer, Sapio

This pull request fixes the V3000 COLLECTION / SGROUP block ordering interoperability issue described in Indigo issue #2109.

Sapio had previously maintained a local hot patch that swapped the emitted SGROUP and COLLECTION sections during CT export, after observing interoperability issues with other toolkits consuming Indigo-generated V3000 output.

Subsequent discussion under the Pistoia Alliance Chemical Exchange Format Committee (CEFC), including clarification from BIOVIA, confirmed that Indigo’s writer should emit these blocks in the documented CTFile order. In particular, the M V30 BEGIN SGROUP ... END SGROUP block should precede the M V30 BEGIN COLLECTION ... END COLLECTION block in V3000 CTAB output. BIOVIA also noted that this ordering helps minimize forward references that must otherwise be maintained by file readers.

This pull request applies that writer-side fix in Indigo and includes unit test coverage for the corrected behavior. Manual validation was also performed after cherry-picking the change into our environment, and the updated output appears to resolve the original interoperability issue.

This fix resolves the observed MOL V3000 interoperability issue with RDKit. At the same time, the CEFC discussion also suggested a broader interoperability principle for V3000 readers: while writers should follow the documented block order, parsers may also benefit from tolerating reasonable block-order variation where practical.

Related to earlier feedback:
[Two V3000 export issues: implict H cause wrong index on s groups, and reversed order for COLLECTON / SGroup Blocks. #2109](https://github.com/epam/Indigo/issues/2109)


core/indigo-core/tests/tests/molecule.cpp
```diff
@@ -29,6 +36,12 @@
 #include <molecule/tpsa.h>
 
 #include "common.h"
+#include "molecule/elements.h"
+#include "molecule/molfile_loader.h"
+#include "molecule/molfile_saver.h"
+#include "reaction/reaction.h"
+#include "reaction/reaction_auto_loader.h"
```
AlexanderSavelyev: why reaction.h in molecule.cpp Please move to reaction tests

AlexanderSavelyev (changes requested):

AlexanderSavelyev:
this change will affect all the tests and the origin issue seems not to be related to the indigo

image

sapiosciences-dev:
> this change will affect all the tests and the origin issue seems not to be related to the indigo
> 
> image

Hi, 

This is the part that was missed.

“Collection Block
[BIOVIA_CTFileFormats_2026.pdf](https://github.com/user-attachments/files/25936764/BIOVIA_CTFileFormats_2026.pdf)

A collection block specifies all collection information for objects in the current connection table context. Collection blocks must be provided after the blocks that define the objects included in the collection to minimize the number of forward object references that must be maintained by the file reader.”

Cited Page 22, in this updated manual provided by BIOVIA in CEFC this year.

sapiosciences-dev:
It is however interesting that in the newest manual, your citation is indeed still there in page 8:
"The atom block and counts line are required. The counts line, atom block, and bond block must appear
in the order indicated. The Sgroup block, 3D block, and link lines can occur in any order after the atom
and bond blocks."

However, I did check with BIOVIA before making the pull request and it is confirmed that the writer fix is correct. 

But I think you also raised a good point that the manual here seems unclear when both sentences appear at same time in version in perspective of a reader/implementer. I'd like to also take your feedback back to BIOVIA team for refinement of this "CT File Formats" manual.

Thanks!

AlexanderSavelyev:
>  Collection blocks must be provided after the blocks that define the objects included

so basically, only if collection includes s-groups, it should be provided after the s-group block

I can't find any information in the specification, that collection should be after sgroups in all cases
I understand, that making collection block last in 100% will be easier, but it will break many tests, which we can't do unfortunately. I would approve the fix for the collection goes after s-group only if s-groups used

sapiosciences-dev:
> > Collection blocks must be provided after the blocks that define the objects included
> 
> so basically, only if collection includes s-groups, it should be provided after the s-group block
> 
> I can't find any information in the specification, that collection should be after sgroups in all cases I understand, that making collection block last in 100% will be easier, but it will break many tests, which we can't do unfortunately. I would approve the fix for the collection goes after s-group only if s-groups used

Yes that should work for us, and it would also be in spirit of the latest CT File Format manual. I can refine the code so that the order is switched only if S-block is used.

sapiosciences-dev:
@AlexanderSavelyev I have made the changes so that S-groups in-use will flip the collection block order, and reverted the logical C++ code order to be the same as before.
I have also adjusted unit tests to have molecule.cpp test file no longer require reaciton and split the current test file into molecules.

It appears there are some tests failing in Python side due to the change of S-group order expectations, though the core tests are now successful.

Do you want to examine those changes yourself first?

At a glance, these are not real failures but script is prescribing these order due to hard coded results. But otherwise satisfies both CT format and parser.

AlexanderSavelyev:
Hi @sapiosciences-dev , we finally started some S-group refactoring. There is a [work-in-progress PR](https://github.com/epam/Indigo/pull/3625) which will change the s-groups properties and also fix the s-group order (right now there is additional bug, that parents with parents can go in wrong order). After we merge the PR, we can return to your PR and even try to update the tests and change the collections order

sapiosciences-dev:
Thank you @AlexanderSavelyev for the quick progress update!

AlexanderSavelyev (approved):

AliaksandrDziarkach:
I add unit test in this PR https://github.com/sapiosciences-dev/Indigo/pull/1

PR #3589, opened 2026-04-13
#3524 – Support default InChI (Key) generation

### LINKED ISSUE #3524: Support default InChI (Key) generation
**Background**  
The current Indigo-InChI support implements some magical adjustments to the options used by the generator. In fact, the code does something like this:

```cpp
   iterate the all stereocenters
       if (type == MoleculeStereocenters::ATOM_AND)
            setOptions("/SRac");
        else if (type == MoleculeStereocenters::ATOM_OR)
            setOptions("/SRel");
```

This setOptions changes the options for all follow-up calls to the InChI with other molecules.

This has two implications:
* Depending on the order of molecules used to call the generator, the result might differ for specific molecules (if you had already called it with some stereo before or not). You can circumvent that by calling resetOptions() before each call.
* There is no chance to really control the generator:
  * No chance to force the generation of standard InChIs.
  * No chance to provide other options.
This is a major problem for us because we are working with molecules from different systems that have already assigned standard InChI keys. There is no way to compare this with the InChI key we have generated using Indigo.

From my understanding, InChI (keys) are designed to be system-independent, and this magical adjustment of the options is exactly the opposite.

**Solution**
Provide a new method (or possibly a parameter to the existing one) to allow for exact control of all options to be used.

**Alternatives**
Add a way to set the options to be used globally.

**Additional context**

Robert-Henning-t2:
## General description of the PR:

Fix for #3524.
Added optional parameter to force InChI generation parameters.

## Generic request
### For release/xx branch
- [ ] backmerge to master (or newer release/xx) branch is created
### Optional
- [x] unit-tests written
- [ ] documentation updated

## Backmerge request
- [ ] PR name follows the pattern `Backmerge: #1234 – issue name`
- [ ] code contains only backmerge changes

## Bump request
- [ ] PR name follows the pattern `Bump version to ...`
- [ ] add brackets \[ \] for 'skip ci' and put it into the body
- [ ] milestone is linked to PR
- [ ] all tickets are closed inside the relevant milestone

### DIFF
--- a/api/c/indigo-inchi/src/indigo_inchi_api.cpp
+++ b/api/c/indigo-inchi/src/indigo_inchi_api.cpp
@@ -152,7 +152,7 @@ CEXPORT int indigoInchiLoadMolecule(const char* inchi_string)
     INDIGO_END(-1);
 }
 
-CEXPORT const char* indigoInchiGetInchi(int molecule, const char* forcedOptions = nullptr)
+CEXPORT const char* indigoInchiGetInchiWithForcedOptions(int molecule, const char* forcedOptions)
 {
     INDIGO_BEGIN
     {
@@ -164,6 +164,11 @@ CEXPORT const char* indigoInchiGetInchi(int molecule, const char* forcedOptions 
         return tmp.string.ptr();
     }
     INDIGO_END(0);
+}
+
+CEXPORT const char* indigoInchiGetInchi(int molecule)
+{
+    return indigoInchiGetInchiWithForcedOptions(molecule, nullptr);
 }
 
 CEXPORT const char* indigoInchiGetInchiKey(const char* inchi_string)
--- a/api/c/indigo-inchi/src/indigo_inchi_api.cpp
+++ b/api/c/indigo-inchi/src/indigo_inchi_api.cpp
@@ -152,7 +152,7 @@ CEXPORT int indigoInchiLoadMolecule(const char* inchi_string)
     INDIGO_END(-1);
 }
 
-CEXPORT const char* indigoInchiGetInchi(int molecule)
+CEXPORT const char* indigoInchiGetInchi(int molecule, const char* forcedOptions)
 {
     INDIGO_BEGIN
     {
@@ -160,7 +160,7 @@ CEXPORT const char* indigoInchiGetInchi(int molecule)
         IndigoObject& obj = self.getObject(molecule);
 
         auto& tmp = self.getThreadTmpData();
-        inchi_wrapper.saveMoleculeIntoInchi(obj.getMolecule(), tmp.string);
+        inchi_wrapper.saveMoleculeIntoInchi(obj.getMolecule(), tmp.string, forcedOptions);
         return tmp.string.ptr();
     }
     INDIGO_END(0);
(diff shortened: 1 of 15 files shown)


BrillBay:
Hi Team Epam/Indigo,

robert made a first shot on #3524. As this 'fix' is really important for us, please have a look at the used approach and the impl itself. If you think we should adjust something, please give us some hints about the direction to go.
Otherwise it will be great if we can have this as soon as possible released.

AlexanderSavelyev:
Hi @Robert-Henning-t2  thanks for the PR. Please fix the static analysis issues. 
one of options can be to run in docker (using existing CI version image) the fix e.g.

`docker run --rm -v $(pwd):/indigo --user $(id -u):$(id -g) epmlsop/indigo-tester:latest sh /indigo/.ci/static_analysis_fix.sh`

Robert-Henning-t2:
@AlexanderSavelyev thanks for the hint, i tried to make the linter/type-checker happy

AlexanderSavelyev:
it seems like segfault in linux tests

`indigo-dlopen-tests-indigo-inchi (SEGFAULT)`

Linux builds are using docker. One can find all scripts in 
https://github.com/epam/Indigo/blob/master/.github/workflows/indigo-ci.yaml

please inform if help required to reproduce

rbri:
@AlexanderSavelyev we will have another look at this tomorrow...

api/c/indigo-inchi/src/indigo_inchi_api.cpp
```diff
@@ -152,7 +152,7 @@ CEXPORT int indigoInchiLoadMolecule(const char* inchi_string)
     INDIGO_END(-1);
 }
 
-CEXPORT const char* indigoInchiGetInchi(int molecule, const char* forcedOptions)
+CEXPORT const char* indigoInchiGetInchi(int molecule, const char* forcedOptions = nullptr)
```
AlexanderSavelyev: C does not support inline parameter initialization. I don't sure `forcedOptions = nullptr` can be used here. The CEXPORT functions go to the dynamic export symbols and called using python, dotnet and java wrappers.

AlexanderSavelyev (changes requested):

api/c/indigo-inchi/src/indigo_inchi_api.cpp
robert-henning: Hi @AlexanderSavelyev,

We understand that default parameter values are a C++ feature and cannot be used in a C-compatible header/export declaration. Before we rework the approach, could you clarify: does the C API/ABI need to remain strictly stable and backward-compatible?

The reason we ask: if backward compatibility must be preserved, we would need to keep the existing indigoInchiGetInchi(int molecule) signature untouched and introduce a new function (e.g. indigoInchiGetInchiWithForcedOptions(int molecul, const char* forcedOptions)) for the extended behavior. If, on the other hand, a breaking change to the C API is acceptable i would only remove the default value and everything should be fine.

AlexanderSavelyev: yes, it is worth to keep back-compatibe c-functions, because someone can use it from c interface. Please create indigoInchiGetInchiWithForcedOptions() and change the usage from different wrappers to this function (for wrappers e.g. python we can specify the default parameter so it also stays back-compatible)

AlexanderSavelyev (approved):

AlexanderSavelyev:
merged, thanks for the PR!

PR #3596, opened 2026-04-14
#3333 implementation to support biovia 2009 and 2017 specification

### LINKED ISSUE #3333: Use BIOVIA post 2014 valence table to calculate atomic valences from molfiles
**Background**
Currently when reading molfiles, indigo uses a pre-2014 biovia specification when calculating valences. This leads to many metal atoms having a default valence which isn't zero and therefore implicit hydrogens are added. In general, a metal atom should have zero valence and zero implicit atoms.

**Solution**
Implement the post-2014 valence table when reading molfiles:
Image

**Additional context**
Different specifications have different valence tables - i.e. molfile, smiles, cdxml. I think this table should only apply to molfiles and maybe include an option to use the old table for backwards compatibility.
[bioviachemicalrepresentation2017.pdf](https://github.com/user-attachments/files/23520110/bioviachemicalrepresentation2017.pdf)

### LINKED ISSUE #3539: Refactor elements and valency algorithm to make it easy to maintenance.
During investigation of the molecular parsing issue #3496. I have noticed that elements class is very complicated and requires introduce new if ... else conditions when it face the specific conditions. After investigation over the certain time how it should be optimized and simplified that it can be easy to read and maintain.

NikolaiBalabanov:
## Generic request
- [x] unit-tests written
- [x] documentation updated


core/indigo-core/molecule/src/elements.cpp
```diff
@@ -273,730 +276,102 @@ int Element::calcValenceOfAromaticAtom(int elem, int charge, int n_arom, int min
...
-                return rad + conn;
-            else if (charge == 2)
-                return rad + conn;
-        }
-    }
-    else if (groupno == 6)
+    // Certain charges are absorbed by electronic structure, not consuming a valence slot
+    bool absorbed = false;
+
+    switch (g)
```
AliaksandrDziarkach: Please rename to group or use Element::group(elem) here

core/indigo-core/molecule/elements.h
```diff
@@ -164,12 +164,15 @@ namespace indigo
...
+    inline bool ElementHydrogenOnLeft(int el)
     {
-        if (el == ELEM_O || el == ELEM_F || el == ELEM_S || el == ELEM_Cl || el == ELEM_Se || el == ELEM_Br || el == ELEM_I)
-            return true;
-        return false;
-    };
+        return el == ELEM_O || el == ELEM_F || el == ELEM_S || el == ELEM_Cl || el == ELEM_Se || el == ELEM_Br || el == ELEM_I;
+    }
+
+    [[deprecated("Misspelled, use ElementHydrogenOnLeft instead")]] inline bool ElementHygrodenOnLeft(int el)
```
AliaksandrDziarkach: Please remove this function - no need to save

api/cpp/tests/rendering/threads.cpp
```diff
@@ -61,8 +61,8 @@ namespace
 TEST(RenderingThreads, Basic)
 {
     std::vector<std::thread> threads;
-    threads.reserve(100);
```
AliaksandrDziarkach, line 64: Please use THREAD_COUNT here too

AliaksandrDziarkach (approved):

PR #3684, opened 2026-05-25
fix SMARTS search in Postgres

### LINKED ISSUE #3683: SMARTS search crashes in the Postgres
**Summary**
SMARTS search is broken on any postgres instance with bingo >= 1.36, when planner tries to search index connection crashes (seq scan works)

**Steps to Reproduce**
1. Run [bingo_bug.sql](https://github.com/user-attachments/files/28212314/bingo_bug.sql) on fresh DB instance with bingo>=1.36 installed

**Actual behavior**
All searches work

**Expected behavior**
SMARTS search gives the result identical to seq scan

**Environment details:**
 - Back-end version >=1.36 and master too

asiomchen:

This patch fixes #3683 - a PostgreSQL regression introduced in PR #3154 for indexed molecule SMARTS searches. The regression came from treating SMARTS like the cursor-backed exact/gross/mass paths and calling `finishOnTransactionEnd()` in `_prepareSmartsSearch()`. 
Unlike those search types, molecule SMARTS reuses the substructure preparation path and never creates `_searchCursor`, so the added deferred-finish call could only fail there. Molecule substructure is unaffected because it calls `_prepareSubSearch()` directly, and reaction SMARTS is unaffected because the analogous reaction code never added the deferred-finish call.

To lock that behavior down, the new pytest coverage builds a dedicated molecule table and forces bingo_idx usage with enable_seqscan = off, then verifies both the plan shape and the returned IDs for exact, substructure, and SMARTS

### DIFF
--- a/bingo/postgres/src/pg_core/mango_pg_search_engine.cpp
+++ b/bingo/postgres/src/pg_core/mango_pg_search_engine.cpp
@@ -481,10 +481,6 @@ void MangoPgSearchEngine::_prepareGrossSearch(PG_OBJECT scan_desc_ptr)
 void MangoPgSearchEngine::_prepareSmartsSearch(PG_OBJECT scan_desc_ptr)
 {
     _prepareSubSearch(scan_desc_ptr);
-    if (_deferred_finish)
-    {
-        _searchCursor->finishOnTransactionEnd();
-    }
 }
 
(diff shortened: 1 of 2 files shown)


AlexanderSavelyev (approved):

AlexanderSavelyev:
merged, thanks for the PR!

PR #3757, opened 2026-06-24
 #3754 - Helm import is case sensitive

### LINKED ISSUE #3754: Helm import is case sensitive
**Steps to Reproduce**
1. Convert two HELM strings to KET:
- RNA1{r(A)p}$$$$V2.0
- RNA1{R(A)P}$$$$V2.0

**Actual behavior**
System recognizes Sugar and Phosphate only for one of strings. For another one indigo convert to unknown monomers.

**Expected behavior**
Helm import should be case insensitive

**Environment details:**
v1.45.0-rc.1

AliaksandrDziarkach:
Fix code. Add UT

## Generic request


.github/workflows/indigo-ci.yaml
```diff
@@ -958,8 +958,8 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJSON(needs.set_matrix.outputs.matrix) }}
     runs-on: ${{ matrix.os }}
-    needs: [build_bingo_postgres_linux_x86_64, build_bingo_postgres_windows_msvc_x86_64, build_bingo_postgres_macos_x86_64, set_matrix]
-    if: ${{ always() && (needs.build_bingo_postgres_macos_x86_64.result == 'success' || needs.build_bingo_postgres_macos_x86_64.result == 'skipped') }}
+    needs: [build_bingo_postgres_linux_x86_64, build_bingo_postgres_windows_msvc_x86_64, build_bingo_postgres_macos_x86_64]
```
NikolaiBalabanov: This job's `strategy.matrix` still reads
`${{ fromJSON(needs.set_matrix.outputs.matrix) }}`, but `set_matrix` was just removed from
`needs`. The `needs` context only contains **direct** dependencies, so `needs.set_matrix` is
null here → `fromJSON(null)` → the matrix fails to evaluate and the job never starts. The run
is marked `failure` even though no individual job shows as failed (confirmed: branch runs
#7507–#7518 are all `failure`, and `build_indigo_utils_x86_64` is absent from the job list).

Every other matrix job in this file keeps `set_matrix` in `needs` (lines 80, 221, 360, 510,
650).
Was it intentionally removed, what was a reason for that?

Image

.github/workflows/indigo-ci.yaml
```diff
@@ -1498,7 +1498,7 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     needs: [build_bingo_postgres_linux_x86_64, build_bingo_postgres_windows_msvc_x86_64, build_bingo_postgres_macos_x86_64]
-    if: ${{ always() && (needs.build_bingo_postgres_macos_x86_64.result == 'success' || needs.build_bingo_postgres_macos_x86_64.result == 'skipped') }}
+    if: ${{ always() && (needs.build_bingo_postgres_linux_x86_64.result == 'success') && (needs.build_bingo_postgres_windows_msvc_x86_64.result == 'success') && (needs.build_bingo_postgres_macos_x86_64.result == 'success' || needs.build_bingo_postgres_macos_x86_64.result == 'skipped') }}
```
NikolaiBalabanov, line 1501: Behavior change worth confirming: the previous condition only checked macOS; now linux and
windows must both be `success`. That means a failed bingo-postgres linux/windows build will now
block `build_indigo_utils_x86_64` and downstream publishing, whereas before they ran
independently. Is this tightening intentional?

core/indigo-core/molecule/src/sequence_loader_helm.cpp
```diff
@@ -40,6 +40,13 @@
 
 using namespace indigo;
 
+std::string to_lower(const std::string& str)
```
NikolaiBalabanov: `to_lower` is a free function with external linkage in the **global** namespace — these risks
ODR collisions across translation units. Also, there's a stray `; ` after the closing brace.

why do we need it in a **global** namespace, maybe it will be better to move into the `indigo` namespace (or an anonymous namespace) and drop the trailing
semicolon? 
There is also idea to use similar approach: `std::transform(..., ::tolower)`; The same pattern already present
in `monomers_template_library.h:79` and `monomer_name_parser.cpp:1679` — what do you think to consider a single shared `inline` helper in `namespace indigo` and reuse it here and in the parts of library lookup functionality?

api/tests/integration/tests/formats/biln.py
```diff
@@ -223,9 +223,6 @@
 
 helm_errors = {
     "CHEM1{[qweqwe]}$$$$V2.0": "Only amino acids and CHEMs with BILN codes can get exported to BILN.",
-    # Unresolved HELM aliases that map to real BILN terminal aliases
-    # must still be validated against the resolved library template attachment points.
-    "PEPTIDE1{[Ac]}|PEPTIDE2{K}$PEPTIDE1,PEPTIDE2,1:R1-1:R3$$$V2.0": "Cannot save in BILN format - unsupported attachment point 'R1'.",
```
NikolaiBalabanov, line 228: These negative tests assert the BILN-export error `unsupported attachment point 'R1'` for
`PEPTIDE1{[Ac]}|PEPTIDE2{K}$...1:R1-1:R3`, which looks unrelated to HELM case-insensitivity.
Why this and another test was removed, is there a certain reason for such actions, could you add a one or few sentence to describe why they no longer apply?

api/tests/integration/ref/formats/biln.py.out
```diff
@@ -38,9 +38,7 @@ helm_underscore_alias:HELM->BILN SUCCEED
 Test 'CHEM1{[qweqwe]}$$$$V2.0': got expected error 'Only amino acids and CHEMs with BILN codes can get exported to BILN.'
 Test 'PEPTIDE1{A}|RNA1{R(A)P}$$$$V2.0': got expected error 'Only amino acids and CHEMs with BILN codes can get exported to BILN.'
 Test 'PEPTIDE1{[Abu].[Sar].[NMeL].V.[NMeL].A.[DAla].[NMeL].[NMeL].[NMeV].[NMeThr4RBut2enyl]}$PEPTIDE1,PEPTIDE1,1:R1-11:R2$$$V2.0': got expected error 'Only amino acids and CHEMs with BILN codes can get exported to BILN.'
-Test 'PEPTIDE1{[Ac]}|PEPTIDE2{K}$PEPTIDE1,PEPTIDE2,1:R1-1:R3$$$V2.0': got expected error 'Cannot save in BILN format - unsupported attachment point 'R1'.'
```
NikolaiBalabanov, line 41: Similar question as in another comment: https://github.com/epam/Indigo/pull/3757/changes#r3491679310

api/tests/integration/tests/formats/helm_case_insensitive.py
```diff
@@ -0,0 +1,74 @@
...
+    indigo.setOption("json-saving-pretty", True)
+
+    ref_path = joinPathPy("ref/", __file__)
+    lib = indigo.loadMonomerLibraryFromFile(
+        os.path.join(ref_path, "monomer_library.ket")
+    )
+
+    test_cases = [
+        # (HELM string, description)
+        ("PEPTIDE1{cya}$$$$V2.0", "helm_cysteic_acid_1"),
```
NikolaiBalabanov: bracket notation `PEPTIDE1{[cya]}` / `{[CYA]}` is missing, is it accidentally missed or another parser conditional logic used for that;
Example: This example taken from one of the old tests, `PEPTIDE1{[Ac]}`

NikolaiBalabanov (changes requested):

api/tests/integration/tests/formats/biln.py
AliaksandrDziarkach: Monomer "Ac-___N-Terminal acetyl" has helm alias "ac" so before this fix this monomer was loaded as unresolved with 4 standard attachment point. Now it resolved and this monomer has only right attachment point so now it fails at helm load with error.

.github/workflows/indigo-ci.yaml
AliaksandrDziarkach: Yes, this is how it should work, it was broken.

.github/workflows/indigo-ci.yaml
AliaksandrDziarkach: by mistake

core/indigo-core/molecule/src/sequence_loader_helm.cpp
AliaksandrDziarkach: I need this function to create a copy of string.
Added static.

NikolaiBalabanov (approved):
All blockers and unresolved issues fixed.
Suggestion was provided regarding shared helper, it's extraction deferred to a separate task.
For reference see comment related to `to_lower` function.

PR #3846, opened 2026-08-13
#3837 Provide a support for Haptic Bond: Introduce models and data structures according to a contract.

### LINKED ISSUE #3837: Introduce datastructure and models that is going to support Haptic Bonds
---

Introduce the core model haptic bonds are built on: an **attachment group** — a set of atoms acting as one collective endpoint of a bond — and the bonds from such a group to atoms outside it.

The group is kept **outside** the molecular graph: making it a vertex would force every algorithm that walks vertices/edges (valence, aromaticity, canonical SMILES, InChI, fingerprints, substructure search) to learn to skip it. An atom-to-atom haptic bond stays an ordinary edge carrying a mark, since that is what it is chemically and it keeps the fragment connected for layout and rendering.

**Done when:** the subsystem exists with its own files; `BaseMolecule` owns it and handles it in `clear()`, merge/submolecule and a `SKIP_*` flag; the group carries the aggregate charge/radical of the pi system; unit tests cover the container contract and prove the isolation invariant — adding a group and a haptic bond changes no atom's valence or implicit hydrogen count.

---

NikolaiBalabanov:
## Generic request
- [x] base branch master is correct
- [x] unit-tests written

### DIFF
--- a/core/indigo-core/molecule/molecule_attachment_groups.h
+++ b/core/indigo-core/molecule/molecule_attachment_groups.h
@@ -1,0 +1,230 @@
+/****************************************************************************
+ * Copyright (C) from 2009 to Present EPAM Systems.
+ *
+ * This file is part of Indigo toolkit.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


core/indigo-core/molecule/molecule_attachment_groups.h
```diff
@@ -0,0 +1,209 @@
...
+        void setRadical(int radical)
+        {
+            _radical = radical;
+        }
+
+    private:
+        void _reset(); // the one place the fields are listed; not virtual — called from the constructor
+
+        std::vector<int> _atoms;
+        std::vector<Bond> _bonds;
```
AlexanderSavelyev: why do we need bonds?

AlexanderSavelyev (changes requested):

AlexanderSavelyev (approved):

Issue #3661, opened 2026-05-15
BILN import/export special cases


even1024:
**Summary**
Fix BILN import/export according to #3541 requirements: aliases with `_`, CHEM monomers with BILN codes, canonical BILN alias output, bond/cycle ordering, and invalid BILN validation.

**Steps to Reproduce**
1. Load generated BILN: `A-1Nal-Cys_Bn-C`
2. Export HELM/KET with library aliases to BILN, e.g. `PEPTIDE1{[Cys_SEt]}$$$$V2.0`
3. Export CHEM monomers to BILN, e.g. `CHEM1{[PEG-2]}$$$$V2.0`
4. Round-trip BILN with cycles and explicit bonds.

**Actual behavior**
- BILN with `_` in monomer alias failed to parse.
- BILN export used HELM aliases in some cases instead of canonical BILN aliases.
- CHEM monomers were rejected even when they had valid BILN codes.
- Some cyclic/explicit bond cases were not canonicalized as required.

**Expected behavior**
- `A-1Nal-Cys_Bn-C` parses successfully.
- HELM alias `[Cys_SEt]` exports as BILN alias `Edc`.
- CHEM with BILN code, e.g. `[PEG-2]`, exports successfully.
- CHEM without BILN code fails with a clear error.
- BILN round-trip produces canonical chain and bond ordering.

**Environment details:**
- Back-end: Indigo
- Front-end/API: indigo-ketcher WASM
- Tests: Python integration and WASM `test.js`

**Attachments**
N/A

**Additional context**
Added regression tests for BILN import/export, CHEM handling, alias canonicalization, invalid BILN strings, large bond ids, and cycle/explicit bond ordering.

Issue #3541, opened 2026-03-13
Support for BILN


ljubica-milovic:
## Background

BILN is a file format similar to HELM. The major difference between it and HELM is the monomer type it supports: only amino acids (type: PEPTIDE) and CHEMs (type: CHEM). The advantage of BILN compared to HELM is its human-readability.

Similarly both to HELM and to SMILES one BILN string will map onto only one chemical structure, but one structure can be represented with multiple strings. To ensure the uniqueness of a BILN string, a set of best practises is defined (described in the export section of this ticket), but strings not constructed using those best practises should also be able to be loaded (described in the import section of this ticket).

### Export of BILN

1. A structure/structures can get exported to BILN only if all the monomers in them are types PEPTIDE and/or CHEM* $\color{Red}{\textbf{and}}$ have BILN codes. If that is not true, and the user tries to save to BILN, Indigo should return an error: "Only amino acids and CHEMs with BILN codes can get exported to BILN."

> '* BILN only supports monomers explicitly defined as CHEMs. The HELM functionality where unknown small molecules get exported as CHEMs with SMILES is non existent in BILN.

$\color{Red}{\textbf{Logic for constructing the BILN string}}$

2. Different backbones are separated from each other with a full stop.

> Same as HELM, a backbone starts with the first monomer who has an occupied R2, bot doesn't have R1 or doesn't have an occupied R1.

2.1. In case of a circular backbone (where all connections are R1-R2), the BILN string that would be the first alphabetically is chosen.

2.2. In case of multiple backbones, they should be ordered by the decreasing number of amino acid monomers, or if equal  by decreasing number of all monomers, or if equal alphabetically.

2.2.1. Every chain with n≤5 monomers in length is considered to be n monomers long irregardless of the amino-acid content. 

3. Monomers in one backbone with default bonds (R1 of one monomer and R2 of another) are written with a hyphen separating them.

4. Monomers are identified by unique codes:
- If a code contains a hyphen it should be written with square brackets around it;
- If a code doesn't contain a hyphen it shouldn't be written with square brackets around it.

5. Non-backbone bonds are written explicitly in round brackets to the right of monomers that participate in the bond with two pieces of information: a bond identifier and the R-group number:
- Bond identifier and the R-group number are separated with a comma, and without spaces.
- A bond identifier is a unique positive integer, and it is 1 for the first explicitly indicated bond in the string, 2 for the second, ...
- The R-group number indicates the number of the R-group that monomer uses to participate in that bond.

[...]

PR #3848, opened 2026-08-17
 #3685 - Support BILN aliases as a monomer property in the library

### LINKED ISSUE #3685: Support BILN aliases as a monomer property in the library (KET & SDF)
**Background**

[Ketcher ticket for introducing BILN](https://github.com/epam/ketcher/issues/9456)
[Indigo ticket for introducing BILN](https://github.com/epam/Indigo/issues/3541)

**Steps to Reproduce**

1. Call convert monomers library to sdf api with next ket:
```
{"root":{"templates":[{"$ref":"monomerTemplate-qwe___qwe"}]},"monomerTemplate-qwe___qwe":{"type":"monomerTemplate","id":"qwe___qwe","class":"CHEM","classHELM":"CHEM","alias":"qwe","fullName":"qwe","naturalAnalogShort":"","modificationTypes":[],"aliasHELM":"","aliasBILN":"bilnalias","atoms":[{"label":"C","location":[-2.165,0.75,0]},{"label":"C","location":[-1.299,0.25,0]},{"label":"C","location":[-0.433,0.75,0]},{"label":"C","location":[0.433,0.25,0]},{"label":"C","location":[1.299,0.75,0]},{"label":"C","location":[2.165,0.25,0]},{"label":"H","location":[0.433,-0.75,0]}],"bonds":[{"type":1,"atoms":[0,1]},{"type":1,"atoms":[1,2]},{"type":1,"atoms":[2,3]},{"type":1,"atoms":[3,4]},{"type":1,"atoms":[4,5]},{"type":1,"atoms":[3,6]}],"attachmentPoints":[{"attachmentAtom":3,"leavingGroup":{"atoms":[6]},"type":"left"}]}}
```

**Actual behavior**

There is no BILN alias in sdf:
```

  -INDIGO-01000000002D

  0  0  0  0  0  0  0  0  0  0  0 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 0 0 0 0 0
M  V30 BEGIN ATOM
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 END BOND
M  V30 END CTAB
M  V30 BEGIN TEMPLATE
M  V30 TEMPLATE 1 LINKER/qwe/qwe FULLNAME=qwe
M  V30 BEGIN CTAB
M  V30 COUNTS 7 6 2 0 0
M  V30 BEGIN ATOM
M  V30 1 C -2.165 0.75 0.0 0
M  V30 2 C -1.299 0.25 0.0 0
M  V30 3 C -0.433 0.75 0.0 0
M  V30 4 C 0.433 0.25 0.0 0
M  V30 5 C 1.299 0.75 0.0 0
M  V30 6 C 2.165 0.25 0.0 0
M  V30 7 H 0.433 -0.75 0.0 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 1 1 2
M  V30 2 1 2 3
M  V30 3 1 3 4
M  V30 4 1 4 5
M  V30 5 1 5 6
M  V30 6 1 4 7
M  V30 END BOND
M  V30 BEGIN SGROUP
M  V30 1 SUP 1 ATOMS=(1 7) XBONDS=(1 6) BRKXYZ=(9 0.000000 0.500000 0.000000 -
M  V30 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000) LABEL=H CLASS=L-
M  V30 GRP
M  V30 2 SUP 2 ATOMS=(6 1 2 3 4 5 6) XBONDS=(1 6) BRKXYZ=(9 0.000000 -0.50000-
M  V30 0 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000) LABE-
M  V30 L=qwe CLASS=LINKER SAP=(3 4 7 Al)
M  V30 END SGROUP
M  V30 END CTAB
M  V30 END TEMPLATE
M  END
>  <type>
monomerTemplate

>  <aliasHELM>


$$$$
```

**Expected behavior**
BILN alias also converted to sdf

**Environment details:**
Indigo 1.44.0-rc.4

AliaksandrDziarkach:
Fix code. 
Update monomer library.
Update UTs.

## Generic request


core/indigo-core/molecule/src/molecule_tgroups.cpp
```diff
@@ -194,6 +194,7 @@ void TGroup::copy_without_fragment(const TGroup& other)
         modification_types.push().copy(other.modification_types[i]);
     }
     aliasAxoLabs.copy(other.aliasAxoLabs);
+    aliasBILN.copy(other.aliasBILN);
```
NikolaiBalabanov, line 198: `copy_without_fragment` was updated but `TGroup::reuse()` (line 62–84) was not — it clears `aliasHELM`, `aliasAxoLabs` and `modification_types`, but not `aliasBILN`.

 `MoleculeTGroups::_tgroups` is a `PtrReusablePool<TGroup>`, which calls `reuse()` when a slot is handed back out (`ptr_reusable_pool.h:174, 207, 515`). So: template A with `aliasBILN: "Bal"` occupies a slot → slot is released → template B *without* an `aliasBILN` lands in it → B is silently written to KET/SDF as `"aliasBILN": "Bal"` and exported to BILN as `Bal`. I think here we may have silent data corruption, what do you think about that?

core/indigo-core/molecule/src/sequence_saver.cpp
```diff
@@ -915,48 +909,23 @@ std::string SequenceSaver::saveBILN(KetDocument& doc)
...
     for (const auto& monomer_id : monomer_ids)
     {
         auto monomer_class = doc.getMonomerClass(monomer_id);
         const auto& monomer = monomers.at(monomer_id);
         if (monomer_class != MonomerClass::AminoAcid && monomer_class != MonomerClass::CHEM)
             throw Error(biln_export_error);
-        auto biln_alias = get_biln_alias(monomer_class, monomer->alias());
         const int node_idx = static_cast<int>(nodes.size());
-        nodes.push_back({monomer_id, monomer->ref(), monomer->alias(), biln_alias.alias, biln_alias.template_ids, monomer_class});
+        auto const& template_id = monomer->templateId();
```
NikolaiBalabanov, line 915: A `templates.at(template_id)` is unguarded. For an ambiguous monomer (a peptide variant such as `(A+G)`), `getMonomerClass()` returns `AminoAcid`, so the class check above passes, but `templateId()` points into `ambiguousTemplates()`, not `templates()` — `std::map::at` throws `std::out_of_range`, which only the generic `catch (const std::exception&)` in `indigo_internal.h:452` picks up. The user gets `map::at` instead of `biln_export_error`.
 Every other site in this file dispatches on `monomerType()` first — lines 352, 516, 1447, 1469. 

Also: `biln_export_error` doesn't name the offending monomer. With the `alias`/`aliasHELM` fallback gone this error will fire far more often (89 BILN-eligible templates in the bundled library have no `aliasBILN`) — please include `monomer->alias()` in the message.

core/indigo-core/molecule/src/sequence_saver.cpp
```diff
@@ -1316,8 +1251,6 @@ std::string SequenceSaver::saveBILN(KetDocument& doc)
             if (chain.monomer_count != 1 || chain.nodes.empty())
                 return 0;
             int node_idx = chain.nodes.front();
-            if (!is_terminal_hyphen_node(node_idx))
```
NikolaiBalabanov, line 1319: This is the change I'd most like justified. The guard is gone:

 ```diff
 -   if (!is_terminal_hyphen_node(node_idx))
 -       return 0;
 ```

 so "a single-monomer chain bound through R2 sorts first, through R1 sorts last" now applies to **every** one-monomer chain, including ordinary amino acids — not just terminal caps. That is what drives the two rewritten expectations in `biln.py`:

 | test | before | after |
 |---|---|---|
 | `biln_multiple_nonbackbone_order` | `A-[Test-6-Ch](1,3)(2,4)-C.D(1,1).E(2,2)` | `E(1,2).A-[Test-6-Ch](1,4)(2,3)-C.D(2,1)` |
 | `biln_equal_chain_topology_order` | `C(1,1).C(1,3)` | `C(1,3).C(1,1)` |

 In the second case there is no canonicalization left at all — output equals input. Bond numbering moved as a side effect, since bond ids are assigned in chain-traversal order.

 Those were the #3661 regression guards for canonical ordering. If the new order is intended, could you state the rule it implements? If it isn't, the guard needs an `aliasBILN`-based replacement (e.g. "single-monomer chain whose template exposes exactly one attachment point") rather than removal. Either way the lambda name is now misleading — nothing about it is terminal- or hyphen-related anymore.

core/indigo-core/molecule/src/sequence_saver.cpp
```diff
@@ -1391,7 +1324,7 @@ std::string SequenceSaver::saveBILN(KetDocument& doc)
             if (monomer_idx > 0)
                 biln_string += '-';
             const int node_idx = chain.nodes[monomer_idx];
-            std::string monomer_text = nodes.at(node_idx).biln_alias;
+            std::string monomer_text = format_biln_alias(nodes.at(node_idx).biln_alias);
```
NikolaiBalabanov, line 1295: With this call site the only one left, `strip_terminal_hyphen` is never passed `true` — the parameter and its `if` body in `format_biln_alias` (line 849) are dead. it's good candidate for removing. Can you please double check the logic that is going to be triggered?

core/indigo-core/molecule/src/monomers_template_library.cpp
```diff
@@ -450,23 +451,14 @@ namespace indigo
         return EMPTY_STRING;
     }
 
-    bool MonomerTemplateLibrary::hasTerminalHyphenAlias(const std::string& monomer_template_id) const
+    const std::string& MonomerTemplateLibrary::getMonomerTemplateIdByAliasBILN(const std::string& alias)
```
NikolaiBalabanov, line 454: No class filter. The scan returns the first match in id order; the caller then checks the class and **throws** if it happens to be a Sugar/Base rather than continuing to look for the AminoAcid/CHEM template with the same code. `getMonomerTemplateIdByAliasHELM` right above filters by class inside the loop — suggest mirroring it:
```cpp
 const std::string& getMonomerTemplateIdByAliasBILN(MonomerClass monomer_class, const std::string& alias);
 ```
 (No nondeterminism here — `_monomer_templates` is a `std::map` — just a wrong-first-hit.)

File: core/indigo-core/molecule/src/monomers_template_library.cpp [hunk 6]
```diff
@@ -637,6 +629,8 @@ namespace indigo
                     {
                         setKetStrProp(mt, aliasHELM, alias_helm);
                     }
+                    if (properties.contains("aliasBILN"))
```
NikolaiBalabanov, line 635: `properties.contains("aliasBILN")` will also set an **empty** string prop when the SDF field is blank; `aliasHELM` two lines up guards with `alias_helm.size()`. `saveBILN` handles the empty case, but the KET output will carry `"aliasBILN": ""`. Suggest matching the `aliasHELM` guard.

api/c/indigo/src/indigo_savers.cpp
```diff
@@ -264,16 +264,20 @@ void IndigoSdfSaver::saveMonomerLibrary(const MonomerTemplateLibrary& monomers_l
...
-        {
-            std::string alias_helm = mt.getStringProp(toUType(MonomerTemplate::StringProps::aliasHELM));
-            auto first = alias_helm.find_first_not_of(" \t\n\r\f\v");
-            if (first == std::string::npos)
-                alias_helm.clear();
-            else
-                alias_helm = alias_helm.substr(first, alias_helm.find_last_not_of(" \t\n\r\f\v") - first + 1);
-            _output.printf(">  <%s>\n%s\n\n", "aliasHELM", alias_helm.c_str());
-        }
+        auto print_alias = [&](MonomerTemplate::StringProps alias_code, const char* alias_str) {
```
NikolaiBalabanov: the lambda is constructed fresh on every iteration of the monomer loop. A file-static helper taking `mt` and `_output` would read better and avoid the per-iteration capture.
What do you think about that?

api/tests/integration/tests/formats/biln.py
```diff
@@ -28,16 +28,16 @@
     # Legacy integration coverage: terminal aliases must keep their
     # explicit non-backbone BILN bonds during import/export roundtrip.
     "biln_terminal_alias": (
```
NikolaiBalabanov, line 28: The comment above `biln_terminal_alias` still reads *"terminal aliases must keep their explicit non-backbone BILN bonds during import/export roundtrip"*, but changes that we have modifies exactly that behavior: `biln_three_chains` collapses from three chains to two, and `helm_terminal_hyphen_left_alias` goes from `A-K(1,2).OMe(1,1)` to `A-K-[-OMe]`. The comment needs to be updated or removed along with the behaviour it documents.

api/tests/integration/tests/formats/biln.py
```diff
@@ -28,16 +28,16 @@
...
         "PEPTIDE1{[ac]}|PEPTIDE2{A.K}$PEPTIDE1,PEPTIDE2,1:R2-2:R3$$$V2.0",
     ),
     "biln_terminal_alias_right": (
         "Bua(1,2).A-K(1,3)",
         "PEPTIDE1{[Bua]}|PEPTIDE2{A.K}$PEPTIDE1,PEPTIDE2,1:R2-2:R3$$$V2.0",
     ),
     "biln_three_chains": (
-        "Ac(1,2).A-K(1,3)(2,2).Me(2,1)",
-        "PEPTIDE1{[ac]}|PEPTIDE2{A.K}|PEPTIDE3{[-Me]}$PEPTIDE1,PEPTIDE2,1:R2-2:R3|PEPTIDE2,PEPTIDE3,2:R2-1:R1$$$V2.0",
+        "ac(1,2).A-K(1,3)(2,2).[-Me](2,1)",
```
NikolaiBalabanov, line 37: Can you please provide information and confirmation to statement:
Is `[-OMe]` valid BILN output? The library gives that template `aliasBILN: "-OMe"`, identical to its KET `alias`. 

Same applies for `-Me`, `-Et`, `-Bn`, `Me-`, `Et-`, `OMe-`, `Bn-`, `NHBn-`. Those look like `alias` copied into `aliasBILN` rather than curated BILN codes, and `format_biln_alias` then brackets them because of the hyphen.

api/tests/integration/tests/formats/idt_unresolved.py
```diff
@@ -22,7 +22,7 @@
 lib = indigo.loadMonomerLibraryFromFile(
     os.path.join(joinPathPy("ref/", __file__), "monomer_library.ket")
 )
-mol = indigo.loadIdt("/i2AmPr/", lib)
+mol = indigo.loadIdt("/i3AmPr/", lib)
```
NikolaiBalabanov, line 25: Are this `/i2AmPr/` → `/i3AmPr/` and other changes including changing in the matching `.out` rewrite are collateral from the library upgrade, unrelated to BILN aliases?

Same for `PEPTIDE3{[Me]}` → `{[meM]}` in `test.js` which is a sequence of library upgrade am I right?

NikolaiBalabanov (changes requested):

api/tests/integration/tests/formats/idt_unresolved.py
AliaksandrDziarkach: Yes. It result of library update.
Old version contains no "2AmPr" template so it cause error, new version contains "2AmPr", so I change it to non-existing 2AmPr to test errors.

core/indigo-core/molecule/src/monomers_template_library.cpp
AliaksandrDziarkach: Only AminoAcid and CHEM can contain alaisBILN

core/indigo-core/molecule/src/sequence_saver.cpp
AliaksandrDziarkach: Miss this part. It should be removed too.

api/tests/integration/tests/formats/biln.py
AliaksandrDziarkach: Yes. [-OMe] is valid.
According to requirements in #3541:

> 4.Monomers are identified by unique codes:

If a code contains a hyphen it should be written with square brackets around it;
If a code doesn't contain a hyphen it shouldn't be written with square brackets around it.

core/indigo-core/molecule/src/sequence_loader_helm.cpp
```diff
@@ -795,10 +747,17 @@ void SequenceLoader::loadBILN(KetDocument& document)
 
             Vec3f monomer_pos(_col * LayoutOptions::DEFAULT_MONOMER_BOND_LENGTH, -LayoutOptions::DEFAULT_MONOMER_BOND_LENGTH * _row, 0);
             ambiguous_template_opts options;
-            const bool has_next_in_chain = data_pos < biln.size() && biln[data_pos] == '-';
-            const auto [monomer_class, monomer_load_alias] = resolve_biln_monomer(monomer_alias, pending_endpoints, has_prev_in_chain, has_next_in_chain);
+            const auto& template_id = _library.getMonomerTemplateIdByAliasBILN(monomer_alias);
```
NikolaiBalabanov, line 750: When loading BILN, `addHelmMonomer` is given the BILN code as the monomer alias, while the
 HELM path (`sequence_loader_helm.cpp:556`) resolves the library `alias` first. In the bundled library 113 of 335 templates have `aliasBILN != alias` (`bAla`/`Bal`, `meG`/`Sar`, `U`/`seC`).
 Is storing the BILN code as the KET monomer alias intentional?

 If yes — what should Ketcher display for such a monomer, and is the mismatch between
 `monomer.alias` and `template.alias` expected downstream? If no — should this mirror the
 HELM path and use the library alias?

core/indigo-core/molecule/src/sequence_loader_helm.cpp
```diff
@@ -795,10 +747,17 @@ void SequenceLoader::loadBILN(KetDocument& document)
...
+            if (template_id.size() == 0)
+                throw_invalid_biln();
+            const auto& monomer_template = _library.getMonomerTemplateById(template_id);
+            auto monomer_class = monomer_template.monomerClass();
+            if (monomer_class != MonomerClass::AminoAcid && monomer_class != MonomerClass::CHEM)
+                throw_invalid_biln();
+            checkAddTemplate(document, monomer_template);
+            _aliasHELM_to_id.emplace(make_pair(monomer_class, to_lower(monomer_alias)), template_id);
```
NikolaiBalabanov: `loadBILN` primes `_aliasHELM_to_id` (keyed on `to_lower(alias)`) so that `addHelmMonomer`
 hits the cache. Since BILN is now case-sensitive, two same-class BILN codes differing only in
 case would collide — `emplace` keeps the first. No such pair exists in the current library.
 Would an `addHelmMonomer` overload taking a resolved `template_id` be preferable to sharing
 the HELM cache?

core/indigo-core/molecule/src/monomers_template_library.cpp
```diff
@@ -450,23 +451,14 @@ namespace indigo
...
-            template_id = getMonomerTemplateIdByAlias(MonomerClass::AminoAcid, monomer_alias + "-");
-        if (template_id.empty())
-            template_id = getMonomerTemplateIdByAlias(MonomerClass::AminoAcid, "-" + monomer_alias);
-        return !template_id.empty() && hasTerminalHyphenAlias(template_id);
+        for (auto& it : _monomer_templates)
+        {
+            if (hasKetStrProp(it.second, aliasBILN) && getKetStrProp(it.second, aliasBILN) == alias)
+                return it.second.id();
+        }
+        return EMPTY_STRING;
```
NikolaiBalabanov, line 464: As it was noted that AminoAcid and CHEM can carry `aliasBILN`, and #3541 req. 1 agrees. But `monomers_template_library.cpp:632` (SDF) and `molecule_json_loader.cpp:1517` (KET) will set it on a template of any class. 

Should the invariant be validated when a library is loaded, so a malformed third-party library fails loudly instead of making the BILN loader throw "invalid BILN string"?

NikolaiBalabanov (changes requested):

core/indigo-core/molecule/src/sequence_loader_helm.cpp
AliaksandrDziarkach: it's ok

core/indigo-core/molecule/src/monomers_template_library.cpp
AliaksandrDziarkach: I think better skip wrong monomer types to be able to export same library as imported

NikolaiBalabanov (approved):

PR #3849, opened 2026-08-19
#3838 introduce to ket format haptic bond support

### LINKED ISSUE #3838: Introduce to KET format haptic bond support
---

Implement the KET format contract agreed with the Ketcher team: `attachmentGroups` declared on a molecule, and a haptic bond expressed as a root-level connection.

Watch out for two contract details: a haptic connection is keyed by **`type`**, while existing monomer connections use `connectionType`; and all ids (`atomId`, `attachmentGroupId`, group `id`) are **strings**, while `attachmentGroups[].atoms` are numbers. On save the molecule is split into components, and a haptic bond crosses that split — the bonds must be collected before splitting.

**Done when:** both reference examples load and survive a round-trip; a group whose molecule node is missing from `root.nodes`, or a connection referring to an unknown group/atom, fails with a clear message; attachment groups feed the connectivity sets used to split a reaction KET, so a ferrocene reactant stays one component instead of three.

---

NikolaiBalabanov:
Update Ket format loader/saver to support new model including Attachment Group.
## Generic request
- [x] unit-tests written

### DIFF
--- a/core/indigo-core/molecule/base_molecule.h
+++ b/core/indigo-core/molecule/base_molecule.h
@@ -122,7 +122,7 @@ namespace indigo
     // merging with molecule and cloning procedures
     enum
     {
-        SKIP_ALL = 0xFF,
+        SKIP_ALL = 0x3FF,
         SKIP_CIS_TRANS = 0x01,
         SKIP_STEREOCENTERS = 0x02,
         SKIP_XYZ = 0x04,
--- a/core/indigo-core/molecule/base_molecule.h
+++ b/core/indigo-core/molecule/base_molecule.h
@@ -122,7 +122,6 @@ namespace indigo
     // merging with molecule and cloning procedures
     enum
     {
-        SKIP_ALL = 0x3FF,
         SKIP_CIS_TRANS = 0x01,
         SKIP_STEREOCENTERS = 0x02,
         SKIP_XYZ = 0x04,
@@ -137,6 +136,10 @@ namespace indigo
         // skipped with them, since a bond without its group is meaningless.
         SKIP_ATTACHMENT_GROUPS = 0x100,
         SKIP_HAPTIC_BONDS = 0x200,
+        // Last on purpose: every flag above has to appear here, so a new one is added
+        // in the line right below itself rather than by recounting a hex literal.
+        SKIP_ALL = SKIP_CIS_TRANS | SKIP_STEREOCENTERS | SKIP_XYZ | SKIP_RGROUP_FRAGMENTS | SKIP_ATTACHMENT_POINTS | SKIP_TGROUPS |
+                   SKIP_TEMPLATE_ATTACHMENT_POINTS | SKIP_RGROUPS | SKIP_ATTACHMENT_GROUPS | SKIP_HAPTIC_BONDS,
     };
 
     class Molecule;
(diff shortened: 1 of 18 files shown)


core/indigo-core/molecule/src/ket_document_json_loader.cpp
```diff
@@ -65,12 +66,17 @@ void KetDocumentJsonLoader::parseConnections(const rapidjson::Value& connections
     for (rapidjson::SizeType i = 0; i < connections.Size(); ++i)
     {
         const auto& connection = connections[i];
-        std::string connection_type = connection["connectionType"].GetString();
+        // A haptic connection carries "type" instead, and belongs to a molecule rather
+        // than to a monomer document. Reading the absent member is UB in release.
+        if (!connection.HasMember(KetConnectionType))
+            throw Error("Connection without \"connectionType\" is not supported in a monomer document");
```
AliaksandrDziarkach: Please use constant for \"connectionType\"

core/indigo-core/molecule/src/ket_document_json_loader.cpp
```diff
@@ -65,12 +66,17 @@ void KetDocumentJsonLoader::parseConnections(const rapidjson::Value& connections
...
+
+        std::string connection_type = connection[KetConnectionType].GetString();
         if (connection_type == KetConnectionSingle || connection_type == KetConnectionHydro)
         {
             KetConnectionEndPoint ep1, ep2;
-            ep1.parseOptsFromKet(connection["endpoint1"]);
-            ep2.parseOptsFromKet(connection["endpoint2"]);
+            ep1.parseOptsFromKet(connection[KetEndpoint1]);
```
AliaksandrDziarkach: Could you change the constant names so that they are different from the class names?
Currently KetEndpoint1 looks like class name e.g. KetConnectionEndPoint .
Maybe something like KetKeyEndpoint1 or something else which will be different from class names.

core/indigo-core/molecule/src/molecule_json_loader.cpp
```diff
@@ -1228,6 +1230,124 @@ void MoleculeJsonLoader::parseSGroups(const rapidjson::Value& sgroups, BaseMolec
...
 
+// Ids of the KET contract are numbers inside strings. Neither obvious route reports
+// a malformed one: atoi() reads "abc" as 0, and the stoi() inside extract_id()
+// throws a std::exception rather than an Indigo one.
+static int parseNumericId(const char* text)
+{
+    if (text == nullptr || *text == '\0')
+        return -1;
+
+    long value = 0;
```
AliaksandrDziarkach: please use long long
In some systems sizeof(int)==sizeof(long) e.g. LLP64 (Win64)

core/indigo-core/molecule/base_molecule.h
```diff
@@ -122,7 +122,7 @@ namespace indigo
     // merging with molecule and cloning procedures
     enum
     {
-        SKIP_ALL = 0xFF,
+        SKIP_ALL = 0x3FF,
```
AliaksandrDziarkach: please move it to the end of enum and add comment that it should be updated when new enum values added

core/indigo-core/molecule/src/ket_document_json_loader.cpp
NikolaiBalabanov: Agree. Fixed.

core/indigo-core/molecule/src/ket_document_json_loader.cpp
NikolaiBalabanov: Correct. Replaced with better naming.

core/indigo-core/molecule/src/molecule_json_loader.cpp
NikolaiBalabanov: @AliaksandrDziarkach, found a better work around, please take a look for that closely.

core/indigo-core/molecule/base_molecule.h
NikolaiBalabanov: Done.

AliaksandrDziarkach (approved):

Issue #2922, opened 2025-05-19
Performing an exact match search results in PostgreSQL termination


amhuhn2:
**Summary**
Performing an exact match search results in PostgreSQL termination

**Steps to Reproduce**
1. RHEL 8.10, PostgreSQL 15, and Bingo bingo-postgres15-linux-x86_64-1.29.0.0.  I have confirmed the same issue exists with every point release between 1.25.0.0 and 1.29.0.0.

2. Add script or SQL to reproduce the issue

```
/* This fails */
SELECT g_0."parent_structure_id"
  FROM "regmol"."parent_structure" AS g_0
 WHERE (g_0."mol_file" @ ($1, '0.1')::bingo.exact) = TRUE;

/* This succeeds with no errors */
SELECT g_0."parent_structure_id"
  FROM "regmol"."parent_structure" AS g_0
 WHERE (g_0."mol_file" @ ($1, '')::bingo.sub) = TRUE;
```

**Actual behavior**
PostgreSQL terminates with the following errors in the log:

(sometimes the following) corrupted size vs. prev_size while consolidating
(OR sometimes the following) double free or corruption (top)
(always the following) ERROR:  error: bingo search engine: unknown index type 0

**Expected behavior**
A result set to be returned, without terminating PostgreSQL.

**Environment details:**
RHEL 8.10, PostgreSQL 15, and Bingo bingo-postgres15-linux-x86_64-1.29.0.0.  I have confirmed the same issue exists with every point release between 1.25.0.0 and 1.29.0.0.

We did not have this issue on our previous environment: CentOS 8.3, PostgreSQL 12, and Bingo 1.9.1.r1

**Attachments**
If applicable, add attachment files to reproduce the issue.

**Additional context**
The error is not 100% consistent.  Also, the following two errors (as reported by bingo.checkmolecule()) exist in our TEST environment (where we are receiving the catastrophic error) but not in our dev environment (where we are NOT experiencing the error):

```
stereocenters: have hydrogen(s) besides implicit hydrogen near stereocenter <number>
molfile loader: empty pseudo-atom
```

rvaidya:
I am also getting this behavior.
PostgreSQL 17 with the bingo binary built from source off of master.

All bingo functions work fine on a toy database with very few records, but with an index with millions of records exact crashes the DB server and sub/sim never complete their queries.

[edit]  See my comment below, slow queries are not a bingo issue.

albertor21:
[...]

Issue #1456, opened 2023-12-22
Bingo in PostgreSQL, make database terminated abnormally easily


Pandaaaa906:
**Steps to Reproduce**
1. PostgreSQL 12.6, compiled by Visual C++ build 1914, 64-bit
2. Bingo 1.9.1.587-g0e3666ae7
3. 32GB Memory just for pg

i got a table "substance"(import from pubchem), more than 40 million records.
bingo will easily make pg break down, even for simple SQL like this:
```SQL
SELECT * FROM "substance"
WHERE iso_smiles @ ('Cc1cc(Cl)nnc1N', 'ALL')::bingo.exact;
```

```text
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

in the pg log:
```text
2023-12-22 09:54:26.261 HKT [19104] LOG:  server process (PID 12960) was terminated by exception 0xC0000005
2023-12-22 09:54:26.261 HKT [19104] DETAIL:  Failed process was running: SELECT * FROM "substance"

	WHERE iso_smiles @ ('Cc1cc(Cl)nnc1N', 'ALL')::bingo.exact
2023-12-22 09:54:26.261 HKT [19104] HINT:  See C include file "ntstatus.h" for a description of the hexadecimal value.
2023-12-22 09:54:26.262 HKT [19104] LOG:  terminating any other active server processes
2023-12-22 09:54:26.266 HKT [19452] WARNING:  terminating connection because of crash of another server process

```

Pandaaaa906:
once the database breakdown
the index usally seem corrupted, and will cause another breakdown...
i have to drop index and recreate index, when the table become massive, it's painful

Pandaaaa906:
Seem to be same and old issue:
[https://github.com/epam/Indigo/issues/108](https://github.com/epam/Indigo/issues/108)
[https://github.com/ggasoftware/indigo/issues/15](https://github.com/ggasoftware/indigo/issues/15)

Pandaaaa906:
@AlexanderSavelyev Any idea?

jank9201:
> once the database breakdown the index usally seem corrupted, and will cause another breakdown...
> 
> ```
> error: bingo buffer: internal error: corrupted block 8148 data len is -8
> ```
> 
> i have to drop index and recreate index, when the table become massive, it's painful

[...]

PR #3857, opened 2026-08-23
#108: Fix Bingo PostgreSQL index corruption during concurrent writes and crash recovery

### LINKED ISSUE #108: Concurrent Bingo inserts result in corruption
When performing inserts from multiple threads concurrently into a column with a Bingo chemical index  the resulting index becomes corrupt.
I attach some code that should quickly elicit the issue. I tested with PSQL9.6 + latest code compiled from source, but the same issue was also reproducible on PSQL9.4 and the last official Bingo release.
[Test.java.txt](https://github.com/epam/Indigo/files/1481933/Test.java.txt)
The postgres driver I used was:
```
    <dependency>
      <groupId>org.postgresql</groupId>
      <artifactId>postgresql</artifactId>
      <version>9.4.1212.jre7</version>
    </dependency>
```

The code initially produces Exceptions with messages like "An I/O error occurred while sending to the backend." followed by exceptions complaining about  corrupted blocks.
If the same code is run with one thread (rather than ten threads) it completes successfully.

This may be the same issue as was reported here:
https://groups.google.com/forum/#!topic/indigo-general/wIfmsla8gXA

rvaidya:
## Summary

Fixes several Bingo PostgreSQL index durability and corruption issues found during concurrent writes, large update workloads, interrupted index builds, and crash recovery.

The main changes are:

- WAL-log Bingo index mutations so committed changes survive PostgreSQL crash recovery.
- Serialize writes to the same Bingo index to avoid concurrent page and metadata corruption.
- Make VACUUM use the same WAL-backed mutation path.
- Finalize `CREATE INDEX` / `REINDEX` explicitly and verify the completed physical index before publishing build metadata.
- Make parallel builds distinguish intentionally rejected structures from unresolved processing failures.
- Correct PostgreSQL reloption handling, including `NTHREADS=-1`.
- Fix repeated `checkmolecule()` calls accumulating backend cursor and callback state.

## Testing

Added PostgreSQL regression coverage for:

- concurrent inserts and non-HOT updates;
- VACUUM and REINDEX;
- parallel builds with `NTHREADS=-1` and explicit worker counts;
- malformed SMILES during parallel builds, including batch boundaries and the final partial batch;
- strict and tolerant invalid-structure handling;
- continued valid inserts and exact searches after rejected structures;
- shadow-table consistency;
- completed indexes containing no all-zero pages;
- repeated `checkmolecule()` calls without unbounded backend memory growth.

The original corruption cases were reproduced against Bingo 1.45 on PostgreSQL 17. The same scenarios no longer reproduce with this patch.

The existing Bingo PostgreSQL functional suite was also run successfully with **34,152 tests passing**.

## Related issues

- Addresses #108
- May also address the corruption / `unknown index type 0` behavior reported in #2922
- May be related to #1456, although that issue was not directly reproduced

## Note

This PR was mostly authored by AI

### DIFF
--- a/bingo/postgres/src/pg_am/pg_bingo_build.cpp
+++ b/bingo/postgres/src/pg_am/pg_bingo_build.cpp
@@ -11,6 +11,7 @@ extern "C"
 #include "catalog/index.h"
 #include "catalog/pg_type.h"
 #include "storage/bufmgr.h"
+#include "utils/memutils.h"
 #include "utils/rel.h"
 
 #if PG_VERSION_NUM / 100 >= 906
@@ -191,6 +192,7 @@ Datum bingo_build(PG_FUNCTION_ARGS)
     //   BlockNumber relpages;
     IndexBuildResult* result = 0;
     double reltuples = 0;
+    ErrorData* volatile build_error = nullptr;
 
     //   signal(SIGINT, &error_handler);
     elog(DEBUG1, "bingo: build: start building index");
@@ -224,10 +226,13 @@ Datum bingo_build(PG_FUNCTION_ARGS)
 
         {
             BingoPgBuild build_engine(index, schema_name, index_schema, true);
+            MemoryContext oldcontext = CurrentMemoryContext;
+
             /*
-             * Do the heap scan and build index
+             * Preserve PostgreSQL errors raised by the heap scan, including
+             * QUERY_CANCELED. Do not convert them into a Bingo XX000 error.
              */
-            BINGO_PG_TRY
+            PG_TRY();
             {
 #if PG_VERSION_NUM / 100 >= 1200
                 reltuples = table_index_build_scan(heap, index, indexInfo, true, true, bingoIndexCallback, (void*)&build_engine, NULL);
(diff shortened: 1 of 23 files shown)


rvaidya:
@AlexanderSavelyev tagging you since you were involved in related issue discussions

rvaidya:
I have most frequently reproduced the errors from:

- When PostgreSQL was crashing as a result of https://github.com/epam/Indigo/issues/2922 which was fixed
- Unclean PostgreSQL stop/termination when dealing with kubernetes node issues
- Mass updates (thousands to millions) to smiles rows in an indexed table

I had ChatGPT/Codex build out test harnesses to reproduce the above cases using Bingo 1.45 and it was consistently able to reproduce errors, which are not reproducible with this set of patches.

Changes were then  rebased on to master and test suites were re-run.

All my testing has been with PostgreSQL 17 in docker

AlexanderSavelyev:
FYI, for format issues, one can use docker image with specific black version

```
docker run --rm \
    -v $INDIGO_DIR:/indigo \
    epmlsop/indigo-tester:latest bash -c "cd /indigo && bash .ci/static_analysis_fix.sh"
```

it can change many files, please only update necessary ones

rvaidya:
@AlexanderSavelyev Ran that command to format the PR changed files

AlexanderSavelyev (approved):

AlexanderSavelyev:
@rvaidya, merged! Much appreciated for the important fix

PR #3864, opened 2026-08-31
 #3834 - expandedMonomersToAtoms throw IndigoException

### LINKED ISSUE #3834: expandedMonomersToAtoms throw IndigoException
**Steps to Reproduce**
Molecule which contains ```Rsp``` cannot be expanded to atoms with Indigo using Ketcher's monomers library:
```
from indigo import Indigo

indigo = Indigo()
library = indigo.loadMonomerLibraryFromFile(str('path/to/monomer_library.ket')

helm = 'PEPTIDE1{[-Et]}|RNA1{[Rsp]}|RNA2{[hn]}$PEPTIDE1,RNA1,1:R1-1:R1|RNA2,RNA1,1:R1-1:R2$$$V2.0'

helm_doc = indigo.loadHelm(helm, library)
mol = indigo.loadMolecule(helm_doc.json())
mol_exp = mol.expandedMonomersToAtoms()
```
because following IndigoException is raised:
IndigoException: stereocenters: Bad bond flipping. Stereocenter pyramid is already full

AD: same exception after expand monomers in ketcher and try to save as MOLv3000

AliaksandrDziarkach:
Fix code.Add UT

## Generic request

### DIFF
--- a/core/indigo-core/molecule/src/molecule_stereocenters.cpp
+++ b/core/indigo-core/molecule/src/molecule_stereocenters.cpp
@@ -1635,7 +1635,7 @@ void MoleculeStereocenters::registerUnfoldedHydrogen(int atom_idx, int added_hyd
     center->pyramid[3] = added_hydrogen;
 }
 
-void MoleculeStereocenters::flipBond(int atom_parent, int atom_from, int atom_to, bool ignore_errors)
+void MoleculeStereocenters::flipBond(int atom_parent, int atom_from, int atom_to)
 {
     if (exists(atom_from))
     {
@@ -1656,7 +1656,7 @@ void MoleculeStereocenters::flipBond(int atom_parent, int atom_from, int atom_to
     {
         _Atom* to_center = _stereocenters.at2(atom_to);
 
-        if (to_center->pyramid[3] != -1 && !ignore_errors)
+        if (to_center->pyramid[3] != -1)
             throw Error("Bad bond flipping. Stereocenter pyramid is already full");
 
         to_center->pyramid[3] = atom_parent;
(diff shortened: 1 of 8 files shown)


core/indigo-core/molecule/src/molecule_stereocenters.cpp
```diff
@@ -1656,7 +1656,7 @@ void MoleculeStereocenters::flipBond(int atom_parent, int atom_from, int atom_to
     {
         _Atom* to_center = _stereocenters.at2(atom_to);
 
-        if (to_center->pyramid[3] != -1)
+        if (to_center->pyramid[3] != -1 && !ignore_errors)
```
NikolaiBalabanov: Why id 3?

core/indigo-core/molecule/src/molecule_stereocenters.cpp
```diff
@@ -1656,7 +1656,7 @@ void MoleculeStereocenters::flipBond(int atom_parent, int atom_from, int atom_to
...
             throw Error("Bad bond flipping. Stereocenter pyramid is already full");
 
         to_center->pyramid[3] = atom_parent;
```
NikolaiBalabanov, line 1662: Here magic number can corrupt structure with different elements.
Can we confirm that we won't replace different element within this function?

NikolaiBalabanov (changes requested):
Provided solution according to analysis doesn't fix the root cause of the issue, it can hide more serious problem that can lead to worst scenarios.

NikolaiBalabanov (approved):
This case and solution are way better. I still have small suggestions; 
I shared within direct messages.

NikolaiBalabanov (approved):
No additional work required.

PR #3867, opened 2026-08-31
#3863 – canonicalSmiles() fails with "SMILES saver: bad cycle number: 100" for highly cyclic valid SMILES

### LINKED ISSUE #3863: canonicalSmiles() fails with "SMILES saver: bad cycle number: 100" for highly cyclic valid SMILES
**Summary**

`canonicalSmiles()` fails with `SMILES saver: bad cycle number: 100` for a valid highly cyclic molecule.

Indigo successfully parses the molecule and can serialize it with `smiles()`. The failure occurs only when generating the canonical SMILES.

The example molecule is PubChem CID 20652954.

**Steps to Reproduce**

1. Install the Indigo Python bindings for Indigo 1.45.0:

```bash
pip install 'epam.indigo==1.45.0'
```

2. Run the following script:

```python
import indigo

SMILES = (
    "C1C2C3C4C5C6C7C8C9C%10CC%11C%10%10C99C88C77C66C55C44C33C22C1"
    "C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55"
    "C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99"
    "C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11"
    "C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10"
    "%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66"
    "C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22"
    "C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44"
    "C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88"
    "C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11"
    "C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10"
    "%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66"
    "C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1CC2C3"
    "C4C5C6C7C8C9C%10C%11"
)

ind = indigo.Indigo()
mol = ind.loadMolecule(SMILES)

print("atoms:", mol.countAtoms())
print("bonds:", mol.countBonds())
print("ordinary smiles:", mol.smiles())
print("canonical smiles:", mol.canonicalSmiles())
```

The molecule loads successfully and `smiles()` also succeeds:

```text
atoms: 231
bonds: 430
ordinary smiles: C1C2C3C4C5C6C7C8C9C%10C%11C%12C%13...
```

`canonicalSmiles()` then fails with:

```text
indigo.indigo.indigo_exception.IndigoException:
SMILES saver: bad cycle number: 100
```

**Actual behavior**

Indigo successfully:

* parses the input SMILES;
* creates a molecule containing 231 atoms and 430 bonds;
* serializes the same molecule with `smiles()`.

Only `canonicalSmiles()` fails.

The ordinary SMILES generated by Indigo uses ring closure numbers well below 100, so the graph itself does not inherently require more than 99 simultaneously available ring closure numbers.

The exception appears to come from the traversal/order selected specifically during canonical SMILES generation. That traversal eventually attempts to allocate ring closure number 100, which `SmilesSaver::_writeCycleNumber()` rejects.

In Indigo 1.45.0:

```cpp
void SmilesSaver::_writeCycleNumber(int n) const
{
    if (n > 0 && n < 10)
        _output.printf("%d", n);
    else if (n >= 10 && n < 100)
        _output.printf("%%%2d", n);
    else
        throw Error("bad cycle number: %d", n);
}
```

The original PubChem SMILES uses only ring closure numbers 1 through 11.

**Expected behavior**

`canonicalSmiles()` should successfully serialize this molecule, since Indigo can parse the structure and serialize the same graph as ordinary SMILES.

Ideally, a fix would preserve the existing canonical SMILES output for molecules that Indigo already canonicalizes successfully. Applications may persist Indigo canonical SMILES and use exact string equality for identity lookups, so changing canonical output for previously supported molecules could be disruptive.

A compatible fix that only changes behavior for structures whose current canonical traversal cannot be serialized would therefore be preferable.

**Environment details:**

* Back-end version: Indigo 1.45.0
* Binding: Python (`epam.indigo==1.45.0`)
* Python: 3.12
* OS: Linux

**Attachments**

No attachment is required; the SMILES in the reproduction above is sufficient.

PubChem CID: `20652954`

**Additional context**

PubChem currently returns identical `SMILES` and `ConnectivitySMILES` values for CID 20652954, so switching between those representations does not avoid the failure.

This appears to be a canonical SMILES serialization issue rather than an invalid-input or general molecule-complexity issue:

```text
loadMolecule()      succeeds
smiles()            succeeds
canonicalSmiles()   fails with "bad cycle number: 100"
```

The same molecule was also tested with RDKit. RDKit successfully parses the structure as 231 atoms and 430 bonds and successfully generates a canonical SMILES. Its canonical serialization uses ring closure labels only up to 56.

That comparison is not intended to suggest using RDKit's canonical SMILES as a substitute for Indigo's. Canonical SMILES strings are toolkit-specific, and applications may depend on Indigo's existing canonical output for exact identity matching. It does, however, show that a deterministic canonical serialization of this graph is possible without exceeding the two-digit ring closure range.

The stronger Indigo-only observation is that `smiles()` can already serialize this exact molecule successfully. The failure is specific to the traversal/order used by `canonicalSmiles()`.

rvaidya:
Fixes #3863.

`canonicalSmiles()` can require more than 99 ring closure labels for highly cyclic molecules, even when the same molecule can be serialized successfully with `smiles()`. `SmilesSaver` previously rejected ring closure numbers above 99, causing canonical SMILES generation to fail with `bad cycle number: 100`.

This change adds support for extended ring closure numbers using the `%(n)` form. Existing output remains unchanged for ring closure numbers below 100:

* `1`–`9` continue to use the single-digit form.
* `10`–`99` continue to use the existing `%nn` form.
* `100` and above use `%(n)`.

The %(n) notation is also supported by RDKit and specified by IUPAC SMILES+ for extended ring closure numbers, rather than being an Indigo-specific representation or a Daylight SMILES representation.
https://www.rdkit.org/docs/RDKit_Book.html#ring-closures

The SMILES loader is updated to read the same extended form so generated SMILES can be parsed and round-tripped correctly.

A regression test covers PubChem CID 20652954 from #3863. It verifies the molecule loads with the expected 231 atoms and 430 bonds, canonical SMILES generation succeeds, the resulting extended ring closure syntax can be parsed again, and both ordinary and canonical SMILES round-trip to the same canonical representation.

Existing canonical SMILES regression tests continue to pass unchanged.

This PR was authored using AI.

## Generic request

* [x] PR name follows the pattern `#1234 – issue name`
* [x] branch name does not contain '#'
* [x] base branch (master or release/xx) is correct
* [x] PR is linked with the issue
* [ ] task status changed to "Code review"
* [x] code follows product standards
* [x] regression tests updated

### Optional

* [ ] unit-tests written
* [ ] documentation updated

### DIFF
--- a/api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
+++ b/api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
@@ -1,0 +1,38 @@
+import os
+import sys
+
+sys.path.append(
+    os.path.normpath(
+        os.path.join(os.path.abspath(__file__), "..", "..", "..", "common")
+    )
+)
+from env_indigo import *  # noqa
+
+indigo = Indigo()
+
+PUBCHEM_CID_20652954_SMILES = (
+    "C1C2C3C4C5C6C7C8C9C%10CC%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99"
+    "C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11"
+    "C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77"
+    "C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22"
+    "C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55"
+    "C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99"
+    "C%10%10C%11C%11C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11"
+    "C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77"
+    "C66C55C44C33C22C1CC2C3C4C5C6C7C8C9C%10C%11"
+)
+
+molecule = indigo.loadMolecule(PUBCHEM_CID_20652954_SMILES)
+assert molecule.countAtoms() == 231
+assert molecule.countBonds() == 430
+
+ordinary = molecule.smiles()
+canonical = molecule.canonicalSmiles()
+
+assert "%(100)" in canonical
+assert indigo.loadMolecule(canonical).canonicalSmiles() == canonical
+assert indigo.loadMolecule(ordinary).canonicalSmiles() == canonical
+
+extended = indigo.loadMolecule("C%(100)CC%(100)")
+assert extended.countAtoms() == 3
+assert extended.countBonds() == 3
--- a/core/indigo-core/molecule/src/smiles_loader_parsers.cpp
+++ b/core/indigo-core/molecule/src/smiles_loader_parsers.cpp
@@ -1169,6 +1169,45 @@ void SmilesLoader::_readOtherStuff()
         _bmol->removeAtoms(to_remove);
 }
 
+int SmilesLoader::_readCycleNumber(int first_char)
+{
+    if (isdigit(first_char))
+    {
+        _scanner.skip(1);
+        return first_char - '0';
+    }
+
+    if (first_char != '%')
+        throw Error("cycle number expected");
+
+    _scanner.skip(1);
+
+    if (_scanner.lookNext() != '(')
+        return _scanner.readIntFix(2);
+
+    _scanner.skip(1);
+
+    int number = 0;
+    int digits = 0;
+
+    while (!_scanner.isEOF() && isdigit(_scanner.lookNext()))
+    {
+        if (digits == 5)
+            throw Error("cycle number is too large");
+
+        number = number * 10 + (_scanner.readChar() - '0');
+        digits++;
+    }
+
+    if (digits == 0 || _scanner.isEOF() || _scanner.readChar() != ')')
+        throw Error("invalid extended cycle number");
+
+    if (number < 1)
+        throw Error("cycle number %d is not allowed", number);
+
+    return number;
+}
+
 void SmilesLoader::_parseMolecule()
 {
     _cycles.clear();
@@ -1192,13 +1231,7 @@ void SmilesLoader::_parseMolecule()
         {
             while (isdigit(next) || next == '%')
             {
-                int number;
-
-                _scanner.skip(1);
-                if (next == '%')
-                    number = _scanner.readIntFix(2);
-                else
-                    number = next - '0';
+                int number = _readCycleNumber(next);
 
                 while (_cycles.size() <= number)
                     _cycles.push().clear();
@@ -1375,13 +1408,7 @@ void SmilesLoader::_parseMolecule()
             {
                 if (isdigit(next) || next == '%')
                 {
-                    int number;
-                    _scanner.skip(1);
-
-                    if (next == '%')
-                        number = _scanner.readIntFix(2);
-                    else
-                        number = next - '0';
+                    int number = _readCycleNumber(next);
 
                     // closing some previous numbered atom, like the last '1' in C1C=CC=CC=1
                     if (number >= 0 && number < _cycles.size() && _cycles[number].beg >= 0)
--- a/core/indigo-core/molecule/smiles_loader.h
+++ b/core/indigo-core/molecule/smiles_loader.h
@@ -176,6 +176,7 @@ namespace indigo
 
         void _loadMolecule();
         void _parseMolecule();
+        int _readCycleNumber(int first_char);
         void _loadParsedMolecule();
         void _validateStereoCenters();
 
--- a/core/indigo-core/molecule/src/smiles_saver.cpp
+++ b/core/indigo-core/molecule/src/smiles_saver.cpp
@@ -757,6 +757,8 @@ void SmilesSaver::_writeCycleNumber(int n) const
         _output.printf("%d", n);
     else if (n >= 10 && n < 100)
         _output.printf("%%%2d", n);
+    else if (n >= 100 && n <= 99999)
+        _output.printf("%%(%d)", n);
     else
         throw Error("bad cycle number: %d", n);
 }


core/indigo-core/molecule/src/smiles_loader_parsers.cpp
```diff
@@ -1169,6 +1169,45 @@ void SmilesLoader::_readOtherStuff()
...
+        _scanner.skip(1);
+        return first_char - '0';
+    }
+
+    if (first_char != '%')
+        throw Error("cycle number expected");
+
+    _scanner.skip(1);
+
+    if (_scanner.lookNext() != '(')
```
AlexanderSavelyev, line 1185: I can't find in daylight SMILES rule for %(...)
https://www.daylight.com/dayhtml/doc/theory/theory.smiles.html

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
```diff
@@ -1169,6 +1169,45 @@ void SmilesLoader::_readOtherStuff()
...
+    while (!_scanner.isEOF() && isdigit(_scanner.lookNext()))
+    {
+        if (digits == 5)
+            throw Error("cycle number is too large");
+
+        number = number * 10 + (_scanner.readChar() - '0');
+        digits++;
+    }
+
+    if (digits == 0 || _scanner.isEOF() || _scanner.readChar() != ')')
```
AlexanderSavelyev, line 1202: no verification for closed bracket. E.g. C%(100C will not throw error

AlexanderSavelyev (changes requested):

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
rvaidya: Daylight SMILES strictly only allows 2 digit numbers for its ring closure labels:

> The ability to re-use ring closure digits makes it possible to specify structures with 10 or more rings. Structures that require more than 10 ring closures to be open at once are exceedingly rare. If necessary or desired, higher-numbered ring closures may be specified by prefacing a two-digit number with percent sign (%). For example, C2%13%24 is a carbon atom with a ring closures 2, 13, and 24 .

https://www.daylight.com/dayhtml/doc/theory/theory.smiles.html

This PR is specifically using the RDKit extension syntax to represent higher digits: https://www.rdkit.org/docs/RDKit_Book.html#ring-closures
This is also part of the IUPAC SMILES+ WIP spec:
https://github.com/IUPAC/IUPAC_SMILES_plus/blob/master/IUPAC_SMILES%2B.asciidoc

Daylight `%##` syntax is still supported.

Note that it is Indigo's canonicalization algorithm that is requiring this extension.  RDKit and other toolkits canonicalize the same SMILES using far fewer labels, while Indigo needs > 100.

It might make sense to disallow this code if SmilesSaver is set to write daylight format - but canonicalSmiles() does not look at the `smiles-saving-format` setting.

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
rvaidya: I tried this locally.
```
import os
import sys

from indigo import Indigo, IndigoException

indigo = Indigo()
smiles = "C%(100C"

print(f"Loading exact SMILES: {smiles!r}")

try:
    indigo.loadMolecule(smiles)
    print("molecule loaded")
    raise SystemExit(1)
except IndigoException as exc:
    print(exc)
```

Indigo 1.45 gives this in a test using your input:
`molecule auto loader: scanner: readIntFix(2): invalid number representation: "(1"`

This PR gives this in a test using your input:
`molecule auto loader: SMILES loader: invalid extended cycle number`

AlexanderSavelyev: @rvaidya  please update ref/ files

`python api/tests/integration/test.py -t 1 -p canonical_smiles_large_ring_numbers -u`

rvaidya: added the generated .out file

api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
```diff
@@ -0,0 +1,38 @@
...
+
+ordinary = molecule.smiles()
+canonical = molecule.canonicalSmiles()
+
+assert "%(100)" in canonical
+assert indigo.loadMolecule(canonical).canonicalSmiles() == canonical
+assert indigo.loadMolecule(ordinary).canonicalSmiles() == canonical
+
+extended = indigo.loadMolecule("C%(100)CC%(100)")
+assert extended.countAtoms() == 3
```
AlexanderSavelyev, line 37: add print(extended.canonicalSmiles()) to update the out file with expected result (not only atoms counts and bond counts)

api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
```diff
@@ -0,0 +1,38 @@
...
+    "C%10%10C99C88C77C66C55C44C33C22C1C1C22C33C44C55C66C77C88C99C%10%10C%11C%11C%10%10C99C88C77"
+    "C66C55C44C33C22C1CC2C3C4C5C6C7C8C9C%10C%11"
+)
+
+molecule = indigo.loadMolecule(PUBCHEM_CID_20652954_SMILES)
+assert molecule.countAtoms() == 231
+assert molecule.countBonds() == 430
+
+ordinary = molecule.smiles()
+canonical = molecule.canonicalSmiles()
```
AlexanderSavelyev, line 30: add the expected result
print(canonical)

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
```diff
@@ -1169,6 +1169,45 @@ void SmilesLoader::_readOtherStuff()
...
+        return _scanner.readIntFix(2);
+
+    _scanner.skip(1);
+
+    int number = 0;
+    int digits = 0;
+
+    while (!_scanner.isEOF() && isdigit(_scanner.lookNext()))
+    {
+        if (digits == 5)
```
AlexanderSavelyev, line 1195: add test for too large cycle 
try
...
except err
print(err)

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
AlexanderSavelyev: add tests for missing ")" and digits=0 in api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py

AlexanderSavelyev (changes requested):

api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
rvaidya: added this print

api/tests/integration/tests/basic/canonical_smiles_large_ring_numbers.py
rvaidya: added this print

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
rvaidya: additional assertions were added for these cases

core/indigo-core/molecule/src/smiles_loader_parsers.cpp
rvaidya: additional assertion was added for this case

AlexanderSavelyev (approved):

AlexanderSavelyev:
@rvaidya merged, thanks so much for the fix

PR #3912, opened 2026-09-10
#3844 haptic bonds layout

### LINKED ISSUE #3844: Provide a support within a layout algorithm for a haptic bond.
---

Need to provide a support for a haptic bond when laying out or cleaning up a structure, a haptic bond between an attachment group and an atom is **1.5× the standard bond length**; an atom-to-atom haptic bond keeps the normal length.

The bond is anchored at the centre of the ring, so the ring radius eats most of its length — at 1× the partner atom ends up sitting on the ring outline, need to provide a correct support to calculate the correct length of the bond.

Note this bond is not a graph edge, so the layout pipeline does not see it through the usual forces. 
Keep the multiplier a parameter rather than a constant.

**Done when: ** layout provides a verified and correct response. The distance from the ligand outlines to the partner atom are correctly calculated and follows the Specification standards and requirements.

---

NikolaiBalabanov:
## Generic request
- [x] task status changed to "In Review"


api/tests/integration/tests/layout/haptic_golden.py
```diff
@@ -0,0 +1,177 @@
...
+#     when that group holds one bond. A chelating or bridging ligand is held by
+#     several at once, and equal lengths and right angles cannot both be had:
+#     the length wins, and the angle is whatever the fit leaves.
+
+BOND_LENGTH = 1.0
+GROUP_MULTIPLIER = 1.5
+TOLERANCE = 0.02
+ANGLE_TOLERANCE = 5.0
+
+indigo = Indigo()
```
Auto-EPM-LSOP-Support, line 30: remove all the code below, use compare_diff, call -u, use json precision 
remove "golden" from name, use something like haptic_layout.py
see the following example
api/tests/integration/tests/formats/mol_to_ket.py

api/tests/integration/tests/layout/haptic_golden_roundtrip.py
```diff
@@ -0,0 +1,138 @@
...
+indigo.setOption("molfile-saving-skip-date", True)
+# V2000 has no way to express a haptic bond, so the V3000 mode is not optional here.
+indigo.setOption("molfile-saving-mode", "3000")
+renderer = IndigoRenderer(indigo)
+indigo.setOption("render-output-format", "svg")
+
+folder = joinPathPy("haptic-golden", __file__)
+
+
+def centroid(points):
```
Auto-EPM-LSOP-Support, line 38: remove all code, use compare_diff and ref files

Auto-EPM-LSOP-Support (changes requested):

api/tests/integration/tests/layout/haptic_golden.py
NikolaiBalabanov: Done.

api/tests/integration/tests/layout/haptic_golden_roundtrip.py
NikolaiBalabanov: Done.

NikolaiBalabanov:
MacOs CI passed: [#7694](https://github.com/epam/Indigo/actions/runs/35471115982)

PR #3913, opened 2026-09-14
#3770 – Add support for custom fonts in SVG and PNG

### LINKED ISSUE #3770: Add support for custom fonts in SVG and PNG
Currently Indigo uses single default font for text objects while saving chemical drawing to SVG or PNG.

Enhance font rendering functionality as follows:
1. Extend Indigo SVG/PNG rendering API with optional parameter -- **font object in TTF format**. If the font is provided, it should be used for rendering texts.
2. For SVG rendering the font characters should be converted to **glyph** objects to ensure rendering and scaling quality.

If bounding box for a text object is specified, apply the following wrapping rules:

1. Wrap the text to the next line if it doesn't fit the width of the bounding box. 
2. If specified bounding box width is less than the width of the first symbol in the rendered text line, the symbol should be rendered ignoring bonding box width.
3. If a new line of text doesn't fit the height of the bounding box and it is not the first line, the rest of text including this line should not be rendered. The top line of text should be rendered even if doesn't fit the height of the bounding box.

**Agreed approach:**

Ketcher sends custom TTF fonts directly in **POST /v2/indigo/render** via **options.fonts** with bundled JSON string.

```
{
  "options": {
    "fonts": "[
      {
        "name": "Arial Bold",
        "data": "<base64>"
      },
      {
        "name": "Arial Italic",
        "data": "<base64>"
      }
    ]"
  }
}
```

**data** contains the Base64-encoded TTF file.
**name** is used only for logging/debugging and does not affect font selection.
Indigo parses the supplied fonts with FreeType and detects their style.
For each KET text part, Indigo selects the first matching font face based on existing **bold / italic** flags.

{
  "text": "test",
  "bold": true,
  "italic": true
}

This should use the first supplied **bold italic** font face.

If no matching custom font is available, Indigo falls back to the existing default rendering.

**Acceptance criteria:**

* Update render() REST API to 
`utils/indigo-service/backend/service/v2/indigo_api.py`

* Update render() WASM 
`api/wasm/indigo-ketcher/indigo-ketcher.cpp`

* Python integration tests updated, indigo-service and wasm tests

konstantin-mordovskih:
## Summary

Adds support for custom static and variable fonts in SVG and PNG rendering.

Custom fonts can be supplied through the `render-fonts` option as a JSON string containing an array of font descriptors with font names and Base64-encoded font data.

## Changes

- Removed `USE_FONT_MANAGER`. Font manager support is included in every renderer build.
- Added `RENDER_ENABLE_NOTO_SANS`, enabled by default, to embed Noto Sans fonts.
- Renamed `RENDER_ENABLE_CJK` to `RENDER_ENABLE_NOTO_SANS_CJK`, disabled by default.
- Noto Sans CJK is embedded only when `RENDER_ENABLE_NOTO_SANS_CJK=ON`; this option also includes Noto Sans.
- Noto font headers are included only when their corresponding build option is enabled.
- Added parsing, validation, Base64 decoding, and in-memory storage for the `render-fonts` renderer option.
- Added support for static and variable fonts with `wght`, `ital`, and `slnt` axes.
- Updated bundled FreeType from 2.12.1 to 2.14.3.
- Added limits for the input JSON, number of fonts, individual font size, and total decoded size.
- Validates SFNT signatures, opens each submitted font with FreeType during `setOption`, and requires a Unicode charmap.
- Keeps decoded font data only. `getOption("render-fonts")` reconstructs Base64 from stored binary data.
- Enables Cairo FreeType support in all standalone renderer builds.
- Added custom font handling to the Indigo Service render API.
- Added integration coverage for:
  - renderer option validation and reset;
  - invalid JSON, Base64, and font data;
  - static fonts;
  - variable fonts;
  - SVG rendering;
  - PNG rendering.
- Added PNG reference images for Linux, macOS, and Windows.

## Font selection

For each `TextItem`, the renderer selects a font in this order:

1. The first custom font with an exact `bold` / `italic` style match that contains all glyphs required by the text.
2. Noto Sans CJK for CJK text when `RENDER_ENABLE_NOTO_SANS_CJK=ON`.
3. Noto Sans when `RENDER_ENABLE_NOTO_SANS=ON` or `RENDER_ENABLE_NOTO_SANS_CJK=ON`.
4. Cairo's default font.

Outside Emscripten, steps 2–3 run only when a custom font was supplied via `render-fonts`; without
one, the renderer falls back to Cairo's platform font (step 4), same as before this PR. Under
Emscripten, where Cairo has no native font backend, Noto Sans is always used regardless of whether
custom fonts are supplied.

The custom font has priority only when it can render the complete text item. If it lacks one or
more glyphs, selection continues through the fallback chain.

For a missing exact custom style, such as Bold text with only a Regular custom font supplied, the
renderer uses the fallback chain.

## Usage

The `options.render-fonts` value is a string containing a serialized JSON array. Each array item must contain a font name and Base64-encoded font data.

Example renderer options:

```json
{
  "render-fonts": "[{\"name\":\"CustomFont-Regular.ttf\",\"data\":\"<Base64-encoded font data>\"}]"
}
```

The bundled string represents the following array:

```json
[
  {
    "name": "CustomFont-Regular.ttf",
    "data": "<Base64-encoded font data>"
  }
]
```

`name` is used for diagnostics. Font selection uses the style metadata extracted from the font file.

## Generic request

- [x] base branch (`master` or `release/xx`) is correct
- [ ] task status changed to "In Review"


api/c/indigo-renderer/src/indigo_render2d.cpp
```diff
@@ -145,6 +155,60 @@ void indigoRenderGetOutputFormat(Array<char>& value)
     value.readString(mode, true);
 }
 
+void indigoRenderSetFonts(const char* fonts)
```
NikolaiBalabanov, line 182: **Blocking — outside WASM, the option is accepted but does nothing, and CI cannot notice.**

`USE_FONT_MANAGER` defaults to `OFF` (`CMakeLists.txt`), is forced `ON` only under `EMSCRIPTEN`
(`CMakeLists.txt`), and no workflow in `.github/` enables it. With it off,
`render_font_face_manager.cpp` is excluded from the build and `RenderContext` discards the fonts with
`(void)fonts;`. This setter still validates, decodes and stores them, so the Python wheel — and
`indigo-service`, which installs that wheel — returns 200 with the default font. That is the agreed
integration path: Ketcher → `POST /v2/indigo/render`.

The test cannot catch this. `custom_fonts.py` reads `os.getenv("USE_FONT_MANAGER", "OFF")`, which
nothing in CI sets, so `test_custom_font_rendering` never runs: the 12 reference PNGs are never
compared, and the SVG and invalid-font checks never execute. `custom_fonts.py.out` is identical in
both modes, so the skip is not visible in the reference either. The REST test sends `"fonts": "[]"`,
so it passes regardless.

api/c/indigo-renderer/src/indigo_render2d.cpp
```diff
@@ -145,6 +155,60 @@ void indigoRenderGetOutputFormat(Array<char>& value)
...
+        if (!font.HasMember("data") || !font["data"].IsString())
+            throw IndigoError("Invalid font at index %u: 'data' must be a Base64 string", i);
+
+        std::string font_name = font["name"].GetString();
+        const Value& font_base64 = font["data"];
+        std::vector<byte> font_data;
+
+        try
+        {
+            font_data = cppcodec::base64_rfc4648::decode<std::vector<byte>>(font_base64.GetString(), font_base64.GetStringLength());
```
NikolaiBalabanov, line 227: **Blocking — untrusted font bytes reach FreeType unchecked.**

Before this PR the vendored FreeType parsed only the built-in Noto fonts. It now parses whatever a
client sends.

* The vendored version is **2.12.1**. CVE-2025-27363 is an out-of-bounds write when parsing TrueType GX / **variable** fonts. It affects
  ≤ 2.13.0, is fixed in 2.13.3 and is listed in CISA KEV. Variable fonts are a headline scenario of
  this PR.
* **No limits.** Nothing caps the number of fonts or their size here, the service accepts a 1 GB body
  (`service/v2/common/config.py`), and nginx sets `client_max_body_size 0`.
* **Retained twice.** `IndigoRenderer::fontsJson` keeps the full Base64 JSON, and
  `RenderParams::fonts` keeps the decoded bytes — about 2.33× the payload, held for the session.
* **Validated late.** This setter checks JSON and Base64 only. A corrupt TTF fails later, inside
  `renderToBuffer`.

Requested:

1. bump the vendored FreeType to ≥ 2.13.3;
2. cap the font count (8 covers the bold/italic matrix), the per-font size and the total size;
3. check the sfnt signature (`0x00010000`, `true`, `ttcf`, `OTTO`) and open each face once here, so
   `setOption` fails fast on a bad font;
4. do not keep the Base64 — have `indigoRenderGetFonts` return names only.

core/render2d/src/render_font_face_manager.cpp
```diff
@@ -28,41 +28,119 @@
...
-            cairo_font_face_destroy(_face_cjk_bold.cairo_face);
-#endif
-#endif
+        if (cairo_face)
+            cairo_font_face_destroy(cairo_face);
     }
 
     cairo_font_face_t* RenderFontFaceManager::selectCairoFontFace(const TextItem& ti)
     {
+        if (auto custom_face = _findFontFace(_custom_faces, ti))
```
NikolaiBalabanov, line 143: **Face selection has two problems. The first one is blocking.**

**CJK regression.** The custom-face lookup runs before the `RENDER_ENABLE_CJK` block, so once any
custom font is supplied, CJK text no longer falls back to NotoSansCJK. A Latin-only font renders
every ideograph as `.notdef`. This can happen today: the CJK WASM bundle is built with
`-DRENDER_ENABLE_CJK=ON` (`indigo-ci.yaml`), and WASM is the one build where the font manager is
on. Please keep the CJK branch first, or check glyph coverage with `FT_Get_Char_Index` before picking a
custom face.

**Partial sets mix typefaces.** `_findFontFace` needs an exact `(bold, italic)` match. With
only a Regular face supplied, bold runs fall back to Noto Bold, so one line shows two families. That
follows the ticket's wording, but users will not expect it. Falling back within the custom set
(Regular as the base, with Cairo synthesising weight or slant) is more predictable.

core/render2d/src/render_font_face_manager.cpp
```diff
@@ -89,53 +167,134 @@ namespace indigo
...
             throw std::runtime_error("error creating cairo font face " + name);
         }
+
+        face.ft_face = free_type_face->face;
+        face.cairo_face = cairo_face;
+        free_type_face.release();
     }
 
-    void RenderFontFaceManager::_loadFontFaces()
+    void RenderFontFaceManager::_loadCustomFontFaces(const PtrArray<RenderFont>& fonts)
```
NikolaiBalabanov, line 275: Three issues in the loading path:

1. **Fonts are re-parsed on every render.** `RenderContext` is built per render call, so this runs
   `FT_Init_FreeType` and reopens every font each time. Each font is also opened once just to probe
   `FT_HAS_MULTIPLE_MASTERS`, then again per variant. A static font costs 2 × `FT_New_Memory_Face`; a
   variable font with `wght` + `ital` costs 5. Four variable fonts add up to about 20 opens per render.
   Reuse the probed face, and ideally cache parsed faces per session keyed by a hash of the bytes.
2. **Likely MSVC build break.** `render2d` builds with `-W4 -WX`, and passing
   `variation_coordinates->size()` (`size_t`) where `FT_UInt` is expected raises C4267, which `/WX`
   turns into an error. CI misses it because only emcc compiles this file. Use
   `static_cast<FT_UInt>(…)`. Please double check this via MSVC.
3. **Dead code.** `Face(const std::string&, const std::shared_ptr<std::vector<byte>>&)` in
   `render_font_face_manager.h` is never called. Please remove it.

api/tests/integration/tests/rendering/custom_fonts.py
```diff
@@ -0,0 +1,268 @@
...
+        if isIronPython():
+            molecule.Dispose()
+
+
+def assert_custom_svg_rendering(indigo, renderer, document, font_sets):
+    default = render(indigo, renderer, document, "svg", "[]")
+    for set_name, _, fonts in font_sets:
+        custom = render(indigo, renderer, document, "svg", fonts)
+        if custom == default:
+            raise AssertionError("%s did not affect SVG rendering" % set_name)
```
NikolaiBalabanov, line 198: **Test coverage** (apart from the CI gating raised in the `indigoRenderSetFonts` thread):

* `custom != default` shows that something changed. It does not show that glyphs were emitted or that
  the right face was used, and no SVG reference is added, although the suite keeps them
  (`ref/linux/*.svg`).
* Not covered: a Regular-only set, two fonts with the same style (the "first match" rule), OTF/WOFF
  input, empty `data`, CJK text with a Latin-only font, an oversized payload.
* No REST test sends a real TTF, and there are no WASM tests. WASM is the one build where the feature
  is active, and `api/wasm/indigo-ketcher/test/` already has `ketcher_text_panel_*_ref.png`
  fixtures to build on.
* Asserting RapidJSON's exact wording (`"Invalid fonts JSON at offset 1: Invalid value."`) makes this
test a hostage to a third-party upgrade. Matching the prefix `"Invalid fonts JSON at offset"` keeps
the intent and drops the coupling. Please review [assertion](https://github.com/epam/Indigo/pull/3913/changes#diff-726b9b453d60680009f8a5c9c364e1652f5c9ab52e5b6b72c1e59a50caceb120R94);

api/c/indigo-renderer/src/indigo_render2d.cpp
```diff
@@ -688,6 +752,7 @@ void IndigoRenderer::setOptionsHandlers()
         mgr->setOptionHandlerInt("render-image-max-height", SETTER_GETTER_INT_OPTION(rp.cnvOpt.maxHeight));
 
         mgr->setOptionHandlerString("render-output-format", indigoRenderSetOutputFormat, indigoRenderGetOutputFormat);
+        mgr->setOptionHandlerString("fonts", indigoRenderSetFonts, indigoRenderGetFonts);
```
NikolaiBalabanov: * `fonts` is the only renderer option without the `render-` prefix, and it shares the session-wide
  option namespace with core options. The ticket spells it `options.fonts`, so this needs agreement,
  but `render-fonts` (keeping `fonts` as an alias if Ketcher already depends on it) would match the
  rest of the API.
* File-scope `using namespace rapidjson;` next to `using namespace indigo;` brings in names
  as generic as `Value`, `Document` and `SizeType`. Qualifying them at the few use sites is safer.

NikolaiBalabanov:
**Requirements 1–3 of this ticket are not implemented, and the ticket does not mark them out of scope.**

The ticket requires the following when a text object has a bounding box:

1. wrap text to the next line when it does not fit the box width;
2. render the first symbol even if the box is narrower than it;
3. skip the line that overflows the box height and everything after it, while always rendering the
   first line.

None of this exists in the renderer. `RenderItemAuxiliary::_drawText`
(`core/render2d/src/render_item_aux.cpp`) uses the bounding-box width only for alignment (centre,
right, justify). Line breaks come only from `line_starts` and explicit `\n`, and nothing clips to
the height. current pull request does not change this code.

Please either provide a confirmation that this is not part of the current scope or why these requirements didn't address within current pull request;

NikolaiBalabanov (changes requested):
Thanks — the FreeType/Cairo lifetime handling is a clear improvement over the previous
`RenderFontFaceManager` (the old `(cairo_destroy_func_t)FT_Done_Face` cast and the leaked
`FT_Library` are both gone), and the variable-font axis handling is well done.

Requesting changes for three blocking issues, each detailed inline:

1. **The feature is compiled out of every non-WASM build and fails silently**, and the rendering
   test never runs in CI, so the green checks do not cover the new code.
2. **Client-supplied font bytes reach a vendored FreeType 2.12.1 with no size or count limits.**
3. **Custom faces override CJK detection**, which breaks ideographs in the shipped CJK WASM bundle.

The other threads (font loading cost, test coverage, naming) are not blocking.

api/c/indigo-renderer/src/indigo_render2d.cpp
AliaksandrDziarkach: ```indigoRenderGetFonts``` could generate valid render-fonts value from binary font data

api/c/indigo-renderer/src/indigo_render2d.cpp
konstantin-mordovskih: Addressed. The renderer option is now named `render-fonts` to follow the existing renderer-option convention. `rapidjson` types are explicitly qualified; the file-level `using namespace rapidjson` was removed.

core/render2d/src/render_font_face_manager.cpp
konstantin-mordovskih: 1. Applied a lightweight fix instead of a cross-render cache: RenderFontFaceManager no longer
   initializes FreeType at all when no custom font was supplied (outside Emscripten) — mirrors the
   selectCairoFontFace() early return, so the common no-custom-font render call pays nothing here
   now, not just the fallback path.

   Left the actual re-parsing-per-render-call for repeated custom-font sets unresolved for now:
   caching parsed faces across renders needs a thread-safety and invalidation design (concurrent
   renders, e.g. indigo-service handling multiple requests), and there's no profiling data yet
   showing this is a real bottleneck versus the layout/rendering work itself. Left a TODO in the
   constructor and would rather track it as a separate follow-up than add that complexity here
   without evidence it's needed.
2. Addressed with an explicit `static_cast<FT_UInt>` for the variation-coordinate count.
3. Removed.

core/render2d/src/render_font_face_manager.cpp
konstantin-mordovskih: 1. Addressed. A custom face is selected only when it has an exact style match and contains all glyphs required by the complete `TextItem`. Otherwise, selection continues to Noto Sans CJK when enabled, then Noto Sans, then Cairo.
2. Kept intentionally. The agreed selection rule requires an exact `bold` / `italic` match. When it is unavailable, the renderer uses the fallback chain instead of synthesising or selecting an approximate custom style.

api/c/indigo-renderer/src/indigo_render2d.cpp
konstantin-mordovskih: 1. Bundled FreeType was updated to 2.14.3.
2. Added limits for JSON input, font count, individual font size, and total decoded font size.
3. `render-fonts` validates the SFNT signature, opens the font with FreeType during `setOption`, and requires a Unicode charmap.
4. The original Base64 JSON is no longer retained. `getOption("render-fonts")` reconstructs Base64 from the stored decoded bytes.

api/c/indigo-renderer/src/indigo_render2d.cpp
konstantin-mordovskih: Addressed, with one refinement: `USE_FONT_MANAGER` was removed and `RenderFontFaceManager` is now
built in every renderer configuration, so `render-fonts` works outside WASM. Outside Emscripten,
though, Noto Sans/CJK is only used when a custom font was actually supplied via `render-fonts`;
without one, the renderer keeps using Cairo's platform font, same as before this PR. Emscripten is
unaffected — no native font backend there, so Noto Sans stays mandatory.

konstantin-mordovskih:
> **Requirements 1–3 of this ticket are not implemented, and the ticket does not mark them out of scope.**
> 
> The ticket requires the following when a text object has a bounding box:
> 
> 1. wrap text to the next line when it does not fit the box width;
> 2. render the first symbol even if the box is narrower than it;
> 3. skip the line that overflows the box height and everything after it, while always rendering the
>    first line.
> 
> None of this exists in the renderer. `RenderItemAuxiliary::_drawText` (`core/render2d/src/render_item_aux.cpp`) uses the bounding-box width only for alignment (centre, right, justify). Line breaks come only from `line_starts` and explicit `\n`, and nothing clips to the height. current pull request does not change this code.
> 
> Please either provide a confirmation that this is not part of the current scope or why these requirements didn't address within current pull request;

As discussed, we're closing this via a dedicated regression test rather than implementing wrap/clip in this PR: **text_bounding_box.py** reproduces the original overflow with the fallback font and shows it fixed once Indigo renders with the same custom font Ketcher used — see **api/tests/integration/tests/rendering/text_bounding_box.py**

api/tests/integration/tests/rendering/custom_fonts.py
konstantin-mordovskih: Addressed:

* Added coverage for a Regular-only set (already in `FONT_SETS`: `Almendra Regular only`), the
  "first match wins" rule for two fonts with the same style (`assert_first_matching_font_wins`),
  real OTF input (`assert_otf_font_accepted`, using `Almendra-Regular.otf` — a genuine
  CFF-flavored OpenType re-encoding of the same Almendra font used elsewhere in this file, not a
  synthetic one), WOFF input (rejected, since `indigoRenderSetFonts` does not recognize the
  `wOFF` signature), and CJK text with a Latin-only custom font
  (`api/wasm/indigo-ketcher/test-cjk/test_cjk.js`: `custom_font_cjk_fallback` — a Latin-only face
  covers none of the CJK glyphs, so the render must be byte-identical to a no-custom-font render).
  Empty `data` and an oversized payload were already covered.
* Added a REST test that sends a real TTF (`indigo_test.py`: `test_render_with_custom_font`) and
  a WASM test against a checked-in PNG reference using a real font (`test.js`: `custom_font`,
  `ketcher_custom_font_test.ket` / `custom_font_ref.png`).
* The RapidJSON offset message is now matched by prefix (`"Invalid fonts JSON at offset 1:"`),
  not the exact third-party wording.

Not changed: no SVG reference was added. `checkImageSimilarity` treats `.svg` as a no-op
(`return ""`, with `# checkSvgSimilarity(filename)` commented out) for the whole integration
suite, not just this test — that stub predates this PR by several years, so an SVG reference
here would not actually assert anything today. Font selection itself is format-independent
(`RenderFontFaceManager::selectCairoFontFace()` has no notion of output format; SVG vs. PNG only
changes which Cairo surface `RenderContext` creates), so the PNG references already added per
font set (`assert_custom_png_rendering`, `ref/{linux,mac,win}/custom_fonts/*.png`) exercise the
same font-selection/glyph-layout path SVG rendering does.

core/render2d/src/render_font_face_manager.cpp
```diff
@@ -18,124 +18,342 @@
...
+        }
+
+        void destroyFreeTypeFace(void* face)
+        {
+            delete static_cast<FreeTypeFace*>(face);
+        }
+
+    } // namespace
+
+    RenderFontFaceManager::RenderFontFaceManager(const std::vector<RenderFont>* fonts) : _fonts(fonts)
```
AliaksandrDziarkach, line 127: please use std::unique_ptr<std::vector<RenderFont>>& insteda of pointer.

core/render2d/render_font_face_manager.h
```diff
@@ -229,33 +230,52 @@ namespace indigo
...
-                key = std::make_unique<cairo_user_data_key_t>();
             }
         };
 
     private:
-        FT_Library _library;
+        std::shared_ptr<FT_LibraryRec_> _library;
+        // Borrowed, outlives this manager. Null: fonts never requested. Non-null (even empty):
+        // explicitly set.
+        const std::vector<RenderFont>* _fonts;
```
AliaksandrDziarkach, line 255: please remove pointer.
use bool flag instead

AliaksandrDziarkach (changes requested):

api/tests/integration/common/rendering/__init__.py
```diff
@@ -157,48 +157,55 @@ def imageDiff(imp1, imp2):
...
-                )
+    dirname = os.path.normpath(
+        os.path.abspath(
+            os.path.join(
+                os.path.dirname(__file__), "..", "..", "tests", "rendering"
             )
         )
+    )
+    try:
+        if shared:
```
AliaksandrDziarkach, line 195: You could use getRefFilepath2 to get system specific ref path

core/render2d/src/render_font_face_manager.cpp
```diff
@@ -18,124 +18,342 @@
...
     {
-        // TODO: fix faces realease
 #ifndef RENDER_EMSCRIPTEN
```
AliaksandrDziarkach: I think this ifndef not needed

AliaksandrDziarkach (changes requested):

core/render2d/src/render_font_face_manager.cpp
konstantin-mordovskih: Addressed

core/render2d/render_font_face_manager.h
konstantin-mordovskih: Addressed

core/render2d/src/render_font_face_manager.cpp
konstantin-mordovskih: It's needed because outside Emscripten this skips FreeType entirely when fonts are explicitly cleared, since the native backend (fontconfig/Quartz/GDI) handles text instead. Emscripten has no native backend, so FreeType must still init there — that's exactly what the guard preserves.

api/tests/integration/common/rendering/__init__.py
konstantin-mordovskih: **getPlatform()** returns "mingw" for a MinGW-built Python, but our reference images live under ref/win/, not ref/mingw/
**_detect_system()** folds that case into "win", which **getRefFilepath2** doesn't do.
Keeping the explicit logic for now; MinGW CI doesn't run this suite yet, so no immediate risk.
Happy to switch once that normalization lives in getPlatform() too.

AliaksandrDziarkach (approved):

NikolaiBalabanov (approved):

PR #3916, opened 2026-09-17
#3915 Allow users to add custom abbreviations to indgio via the API

### LINKED ISSUE #3915: Allow users to add custom abbreviations to indgio via the API
**Background**
Indigo uses abbreviations.xml to store a list of chemical abbreviations. However, it is missing some common ones and users often have new and esoteric abbreviations that they want to use. In order to add new abbreviations, the user must edit abbreviations.xml, then recompile Indigo.

**Solution**
Allow the user to load their own abbreviations file (xml) via the various APIs. This way indigo doesn't have to be recompiled, a tracked file doesn't have to be changed, and indigo doesn't have to keep updating abbreviations.xml with new abbreviations.

**Alternatives**
Add more abbreviations to abbreviations.xml, however each user or lab will have their custom abbreviations. Handling all these is impossible.

jblack-mestre:
## Generic request
- [ ] task status changed to "In Review"


api/c/indigo/src/indigo_abbreviations_core.cpp
AliaksandrDziarkach, line 56: I think it should be just 
```
        void IndigoAbbreviations::loadDefault()
        {
                loadFromXmlString(default_abbreviations_xml);
        }
```

AliaksandrDziarkach (approved):

